PR #1, opened 2024-05-30
extract iso and replace dol on build

jonathonbrady:
build script quality of life improvements:
- requires iso to be present
- installs dependencies (ninja, nod)
- enables `--no-check` flag on build to ignore sha1 mismatches
- extracts and patches iso with patched dol from build

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,9 +1,19 @@
+import glob
 import os
+import requests
 import shutil
 import subprocess
+import zipfile
+from io import BytesIO
 
 DECOMP_ROOT = os.path.join(os.getcwd(), 'pikmin2/src')
 P2GZ_ROOT = os.path.join(os.getcwd(), 'src')
+
+try:
+    iso = glob.glob(os.path.join(os.getcwd(), '*.iso'))[0]
+except IndexError:
+    print('No .iso file found in the current directory')
+    exit()
 
 for p2gz_root, _, _ in os.walk(P2GZ_ROOT):
     decomp_root = p2gz_root.replace(P2GZ_ROOT, DECOMP_ROOT)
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -28,5 +38,20 @@
             shutil.copy2(p2gz_file, decomp_file)
             print(f'Replaced {decomp_file} with {p2gz_file}')
 
-subprocess.run('python configure.py', cwd='pikmin2')
-subprocess.run('ninja', cwd='pikmin2')
\ No newline at end of file
+if not os.path.exists(os.path.join(os.getcwd(), 'ninja.exe')):
+    ninja_zip = requests.get('https://github.com/ninja-build/ninja/releases/download/v1.12.1/ninja-win.zip').content
+    with zipfile.ZipFile(BytesIO(ninja_zip)) as f:
+        f.extractall()
```
mayabyte, line 44: Windows specific build steps >:U

build.py
```diff
@@ -28,5 +38,20 @@
...
+
+subprocess.run('python configure.py --no-check', cwd='pikmin2')
+subprocess.run('ninja', cwd='pikmin2')
+
+if not os.path.exists(os.path.join(os.getcwd(), 'nodtool.exe')):
+    nodtool = requests.get('https://github.com/AxioDL/nod/releases/download/v1.0/nodtool.v1.win64.exe')
```
mayabyte, line 50: Windows specific build steps part 2 >:U

mayabyte (changes requested):
windows is a plague
installing dependencies automatically would be nice but tbh I'd rather we just tell people to install them in a README or something and assume the binaries exist on path for the sake of the build script

jonathonbrady:
I'll just remove dependency installation for now since no one publicly is building the hack any time soon. I still think we should support cross-platform dependency installation in the future to make installing the hack as painless as possible when the time comes.

mayabyte (approved):
you're the goat. build will be made idiot proof later 👍

PR #4, opened 2024-05-30
Add basic README with build instructions

mayabyte:


HeartPiece44 (approved):
looks schmick to me

jonathonbrady (changes requested):
Can you add a link to at least nodtool since it's kind of obscure?

PR #5, opened 2024-05-31
warn when attempting to patch an unlinked file

jonathonbrady:
Adds a sanity check to warn users if trying to patch an unlinked file by reading from each library's Makefile. I deemed this less bad than trying to import and parse the static array in `configure.py`. We can revisit this later if decomp officially stops supporting make.

### DIFF
--- a/build.py
+++ b/build.py
@@ -4,8 +4,8 @@ import re
 import shutil
 import subprocess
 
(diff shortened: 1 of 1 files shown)


mayabyte (approved):
dank

PR #6, opened 2024-05-31
delete thp files after extracting iso

jonathonbrady:
Deletes all pre-rendered cutscenes from `root/files/thp` on build and patches `sysGCU/pikmin2THPPlayer.cpp` by removing the function body of `THPPlayer::load(EMovieIndex)` and the static array `sTHPPlayerFileSettingTable`. The game still transitions to the cutscene playing state on the start menu but nothing loads. This saves 700 MB.

### DIFF
--- a/build.py
+++ b/build.py
@@ -50,7 +50,7 @@ subprocess.run('ninja', cwd='pikmin2', shell=True)
 if not os.path.exists(os.path.join(os.getcwd(), 'root')):
     subprocess.run(f'nodtool extract "{iso}" root', shell=True)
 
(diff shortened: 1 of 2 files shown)


build.py
```diff
@@ -50,4 +50,7 @@
 if not os.path.exists(os.path.join(os.getcwd(), 'root')):
     subprocess.run(f'nodtool extract "{iso}" root', shell=True)
 
+    for file in glob.glob(os.path.join('pikmin2/root/files/thp', '*.thp')):
```
mayabyte, line 53: shouldn't this path just be `root/files/thp`?

jonathonbrady: oops yes

mayabyte (approved):

PR #9, opened 2024-06-03
patch szs archives and unlinked files on build

jonathonbrady:
Pretty much a complete rewrite of `build.py`.
1. Links any unlinked file that has a patched version.
2. Traverses `files` and `root/files` looking for folders in the former that match the name of szs archives in the latter. If there is a match, the contents of the folder are copied into the archive, and any json files are automatically encoded as bmg. We will add support for other file types later.
3. Requires that new `src` files are in `p2gz` and creates a separate library for them in `libs.json`.

Unfortunately this adds three new dependencies which we will remove later after the release of `cube`. The szs patching code is also hopefully significantly more involved than what will be needed with `cube`.

I tested this by building the iso with a modified szs archive file at `files/messages/mesRes_eng/pikmin2.json`, a dummy file at `src/p2gz/test.cpp`, and an unlinked file at `src/plugProjectNishimuraU/ElecHiba.cpp`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 *.exe
 *.iso
 root
+pikminBMGtool.py
--- a/README.md
+++ b/README.md
@@ -7,8 +7,10 @@
 ### Building
 
 Prerequisites:
+1. [Lunaboy's RARC Tools](https://www.romhacking.net/utilities/1024/)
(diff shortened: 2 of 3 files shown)


HeartPiece44 (approved):
ballin

PR #10, opened 2024-06-09
add menu for editing sprays and pokos

jonathonbrady:
Adds a new menu that edits the current spray and Poko count, unlocking sprays if the demo flag is not set. Counters are validated such that negative numbers are not allowed, spray counts above 99 are not allowed, and Poko counts above 99,999 are not allowed.

The basic implementation is every menu option is a struct that inherits from `MenuOption` that requires a `J2DTextBoxEx*` and `CallBack_CounterRV*`, as well as any overrides for `up()`, `down()`, `left()`, and `right()` to define what control stick directions do when the option is being edited (default behavior is do nothing).

The menu doesn't look good, of course, but there are still many decisions left to be made about how we want to organize all options across all menus, as well as how the code is structured for each menu, so I didn't put much effort into visuals.

Annotation convention I used: `@P2GZ` above entirely new functions/structs/etc, `@P2GZ start` and `@P2GZ end` around multi-line modifications to existing functions/structs/etc, and `@P2GZ` on the same line for single-line modifications.

Also the work to actually enable the menu and updating the README was stolen from HeartPiece's testing branch, thanks for doing that.


HeartPiece44 (changes requested):
thanks for implementing so much of this! this is sick. I think a couple tweaks are needed for this to be fully functional:
- currently, adjusting the pokos over another percent level (so say, 3254 pokos -> 4001 pokos) causes the game to freeze gameplay as it's not correctly handling the percentage cutscenes

there are also a couple QoL changes that would be nice but are less essential
- pressing A when done setting a value should probably also accept the value, rather than just B at present
- would it be possible to add some visual change to the poko digit you're changing to show it's the one being changed? at the moment it's hard to tell which digit is being adjusted without actually adjusting it

jonathonbrady:
1. Are you referring to `playData->mDebtProgressFlags`? I don't see anywhere where the game actually sets these flags.
2. Sure, that's a simple change.
3. I agree. I tried passing in an array of `J2DPictureEx*` to `PokoCountMenuOption` and changing the alpha in `left()` and `right()` but it didn't seem to do anything. I also didn't see a `setColor()` method in `J2DPictureEx` or something like that. Open to ideas on how to make it more clear since many options will have multiple digits.

HeartPiece44:
1. I think the flag setting for `playData->mDebtProgressFlags` happens during some `GameState` functions - the main things to hook into will be via `static_cast<Game::SingleGame::GameState*>(static_cast<Game::SingleGameSection*>(Game::gameSystem->mSection)->mCurrentState)`, either by manually calling some of the `needRepayDemo` or `startRepayDemo` functions, or by adding a block to `GameState::onMovieDone` in `plugProjectKandoU/singleGS_MainGame.cpp` to handle returning from the pause menu. Might be worthwhile adding a flag to something that gets set if pokos update, and then only checking if that flag is set (+ resetting it)? Unsure.
2. ezpz
3. yeah my suggestion would've been just adjusting the alphas on the other values - I'm not sure why that doesn't work honestly, there's probably some other alpha value that needs adjusting internally on the digits themselves, but I'm not entirely sure where that is atm. probably outside the scope of this PR if it doesn't immediately jump out, so dw too much

jonathonbrady:
A lot of changes between the previous commit and these:
1. Fix percent cutscene freeze by returning `false` from `GameState::needRepayDemo()` if the game is paused, which includes being in a menu.
2. Allow saving the current value with the A button.
3. Remove line in `CallBack_CounterRV::setKetaSub` that forces digits to inherit the parent pane's alpha. This required bumping the decomp submodule commit to use an equivalent version. Thanks for doing that so quickly.
4. Completely rework how `MenuOption`s work since there was a lot of duplicate logic in every derived struct. I tried to make the base virtual functions as generic and comprehensive as possible to streamline adding new options. Let me know if it looks reasonable.

HeartPiece44:
Nice! Appreciate you making all these additions so quickly, it's shaping up really well. Only issue I can see now is that some counters (active squad count during gameplay, play time on file select) have some digits not displaying correctly. I assume it's to do with the edits to `CallBack_CounterRV::setKetaSub` but I'm honestly not sure why that's happening. If we can find out why or just patch a workaround that'd be fab.

jonathonbrady:
Introduced a flag in `CallBack_CounterRV` that skips setting digit alpha to parent pane alpha if true.

HeartPiece44 (approved):
Nice work - looks great to me, should make setting up new menu options super easy and nice.

PR #11, opened 2024-06-09
add `--clean` flag to build script

jonathonbrady:
Adds a `--clean` flag to the build script to re-extract the ISO and reset the `pikmin2` submodule before building.


HeartPiece44 (approved):
hawt, looks clean af to me

PR #13, opened 2024-07-08
Use cube for build script

mayabyte:
Had to change the extracted BMG to use cube's format. Otherwise seems to work pretty well


build.py
```diff
@@ -121,41 +121,22 @@
         # patching existing asset
         if os.path.exists(archive):
             existing_directories = set(next(os.walk(iso_path))[1])
-            subprocess.run(f'ArcExtract {archive}', shell=True, stdout=open(os.devnull, 'w'))
+            subprocess.run(f'cube extract {archive}', shell=True)
             extracted_archive = os.path.join(iso_path, (set(next(os.walk(iso_path))[1]) - existing_directories).pop())
```
jonathonbrady, line 125: Minor thing but if cube allows specifying the name of the output folder I would prefer to do that and remove both of the surrounding lines here since they were only necessary for ArcExtract.

mayabyte: it does yeah, let me give that a shot

jonathonbrady:
Is ISO extraction functional in cube? We could remove the dependency on nodtool while we're here.

mayabyte:
It is, but it's a lot slower than nodtool (and caveripper for that matter, which is weird because they use the same library). I haven't had the chance to optimise that yet so I'd prefer to switch it over later on.

jonathonbrady:
Ran `python3 build.py --clean` and am getting the following crash when trying to load one of the custom menus:
oks71OP

mayabyte:
fuck!!

mayabyte:
Cube 0.4.7 addresses the problem 👍

build.py
```diff
@@ -120,42 +120,21 @@
...
         elif patch_dir in P2GZ_CUSTOM_ASSETS:
             print(f'Copying {patch_dir} to {iso_dir}')
             shutil.copytree(patch_dir, iso_dir, dirs_exist_ok=True,
                             ignore=lambda _, contents: [file for file in contents if file.endswith('json')])
             
-            for root, _, files in os.walk(patch_dir):
-                for file in files:
-                    if file.endswith('json'):
-                        subprocess.run(f'python3 pikminBMGtool.py PACK {os.path.join(patch_dir, file)} \
-                                       {os.path.join(extracted_archive, file.replace("json", "bmg"))}', shell=True, stdout=open(os.devnull, 'w'))
```
jonathonbrady, line 152: Pulled the latest and ran `python3 build.py --clean`. I think the build script is just extracting and repacking `mesRes_eng` without packing `pikmin2.bmg.json`, as none of the custom messages show up:

jonathonbrady (changes requested):

build.py
mayabyte: ah, wasn't sure what that was supposed to look like. seems like it was somehow an issue with the copytree call, fixed now

jonathonbrady (dismissed):

mayabyte:
Whoops lol, idk how I missed that. Fixed now

jonathonbrady (approved):

PR #14, opened 2024-07-14
Early Blues Patch

mayabyte:
It Patches Early Blues .


HeartPiece44 (approved):
Awesome!! Thanks for implementing this - looks nice and neat. I've given it a stress test with a bunch of EB attempts and zero softlocks, the rest of the behaviour seems as normal, so I think it's good to go!

PR #16, opened 2024-07-20
add warp menu and teleport anywhere

jonathonbrady:
- Adds a primitive warp menu to the pause screen, allowing warping to any above ground area or sublevel. Only the current squad is warped - no presets or editing yet.
- Press d-pad left to open the teleport anywhere feature. I think this is a good thing to have as a button combo rather than in a menu. Supports immediate warping, switching captains, and saving and reloading 4 different positions.

Almost certainly some bug somewhere but I tried to test the obvious stuff. Also there are a few huge functions but I want to work on injection next so I'll refactor them then.


src/plugProjectNishimuraU/playCamera.cpp
```diff
@@ -0,0 +1,1283 @@
...
+	if (flags & CAMFLAGS_CenterBehind) {
+		setFollowTime();
+	}
+	if (flags & CAMFLAGS_SmoothFollow) {
+		setSmoothThetaSpeed();
+	}
+	changeTargetTheta();
+    if (!p2gz->mIsScrollingCamera) {
+        changeTargetAtPosition();
+    }
```
mayabyte, line 205: needs a P2GZ comment

mayabyte (changes requested):
Works really really well! There are two functionality issues that I think should be addressed before merging:
1. When warping, only the pikmin that are actually following your captain are warped with you. I think this should warp all (non-wild ofc) pikmin that are on the field instead
2. The warp menu isn't accessible in caves, I guess because the squad editor isn't accessible there either

Also, ideally I'd like the menu to remember your selection and keep it there next time you open it, but it's chill to do that another time if you just wanna get this merged

jonathonbrady:
> Works really really well! There are two functionality issues that I think should be addressed before merging:
> 
>     1. ~When warping, only the pikmin that are actually following your captain are warped with you. I think this should warp all (non-wild ofc) pikmin that are on the field instead~ actually tbh ignore this one, it won't matter when we do squad presets anyway
> 
>     2. The warp menu isn't accessible in caves, I guess because the squad editor isn't accessible there either
> 
> 
> Also, ideally I'd like the menu to remember your selection and keep it there next time you open it, but it's chill to do that another time if you just wanna get this merged

1. Only warping the pikmin following the active captain was intentional for when we add squad presets and players _don't_ want to use a preset, since we'll (hopefully) repurpose the squad menu for that and it's easier to just make that the default behavior for consistency
2. Done, we needed to edit ogObjSMenuPauseDoukutu.cpp and ogSceneSMenuPauseDoukutu.cpp
3. Done

PR #18, opened 2024-07-22
In-game timer

mayabyte:
image

First one is for the full cave, second one is for the current sublevel. ~~I don't believe this is fully RTA equivalent yet - it seems to stop briefly between sublevels, so that's something we can improve in the future.~~ It's RTA accurate now, except for the start and end points. I consider this acceptable.

Also works above-ground, but only shows one timer obviously.

### DIFF
--- a/src/plugProjectKandoU/singleGS_CaveGame.cpp
+++ b/src/plugProjectKandoU/singleGS_CaveGame.cpp
@@ -31,6 +31,8 @@
 #include "Game/Entities/ItemPikihead.h"
 #include "nans.h"
 #include "GlobalData.h" // @P2GZ
+#include "JSystem/J2D/J2DPrint.h" // @P2GZ
+#include "P2JME/P2JME.h" //@P2GZ
 
 static const u32 padding[]    = { 0, 0, 0 };
 static const char className[] = "SingleGS_Game";
@@ -230,13 +232,6 @@ void CaveState::exec(SingleGameSection* game)
 	}
 	// @P2GZ End
 
-	// @P2GZ Start - In-game timer
-	p2gz->history->peek()->timer += sys->getDeltaTime();
-	if (p2gz->showTimer) {
-		p2gz->drawTimer();
-	}
-	// @P2GZ End
-
 	// the saving between cave floors is part of this state
 	if (mDrawSave) {
 		particle2dMgr->update();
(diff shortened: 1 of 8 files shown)


src/plugProjectKandoU/singleGS_CaveGame.cpp
```diff
@@ -297,12 +299,45 @@ void CaveState::exec(SingleGameSection* game)
...
  * @note Address: 0x80217D44
  * @note Size: 0x98
  */
 void CaveState::draw(SingleGameSection* game, Graphics& gfx)
 {
+	// @P2GZ Start - timer
+	f32 dt = sys->getDeltaTime();
+	p2gz->history->peek()->timer += dt;
+	mCaveTimer += dt;
+	// @P2GZ End
```
jonathonbrady, line 339: I'm assuming the culprit for the timer pausing between sublevels is `CaveState::draw` just doesn't run when there's nothing to draw. Can we check if `CaveState::exec` or some other exec function still runs every frame between sublevels, and if so, add the delta time since the previous invocation of `CaveState::draw` to `dt` on the next invocation? (Or just move this to `CaveState::exec`?)

src/plugProjectKandoU/singleGS_MainGame.cpp
```diff
@@ -967,12 +976,39 @@ void GameState::drawRepayDemo(Graphics&)
...
 /**
  * @note Address: 0x802174B8
  * @note Size: 0x78
  */
 void GameState::draw(SingleGameSection* game, Graphics& gfx)
 {
+	// @P2GZ Start - timer
+	f32 dt = sys->getDeltaTime();
+	p2gz->history->peek()->timer += dt;
+	// @P2GZ End
```
jonathonbrady, line 1010: Likewise for this - are there weird `GameState::draw` pauses with, e.g., cutscenes above ground?

src/plugProjectKandoU/singleGS_CaveGame.cpp
mayabyte: `exec` has the same problem unfortunately. I think the ideal place to put this would be somewhere higher up in the tree, but I'm not completely sure where that would be yet.

src/plugProjectKandoU/singleGS_MainGame.cpp
mayabyte: `draw` pretty much always runs except during loads as far as I can tell, since cutscenes are still part of the same game state

src/plugProjectKandoU/singleGS_CaveGame.cpp
mayabyte: actually, I have another idea I can try that I scrapped earlier because I thought it was crashing the game, but the problem was something else. lemme give that a shot first

src/plugProjectKandoU/singleGS_MainGame.cpp
```diff
@@ -1289,6 +1306,29 @@ void GameState::drawRepayDemo(Graphics&)
...
+    caveTimerText.initiate();
+    caveTimerText.mCharColor.set(JUtility::TColor(255, 255, 255, 128));
+    caveTimerText.mGradientColor.set(JUtility::TColor(255, 255, 255, 128));
+    caveTimerText.mGlyphWidth = 16.0f;
+    caveTimerText.mGlyphHeight = 16.0f;
+
+	s64 minutes = timerMs / (60 * 1000);
+	s64 seconds = (timerMs / 1000) % 60;
+	s64 tenths = (timerMs / 100) % 10;
+    caveTimerText.print(16, 16, "%lld:%.2lld.%.1lld", minutes, seconds, tenths);
```
jonathonbrady, line 1329: nit: rename `caveTimerText` (I'm assuming this was copy-pasted from `CaveState::drawTimer`)

jonathonbrady (approved):

src/plugProjectKandoU/singleGS_MainGame.cpp
mayabyte: right on the money. fixes

PR #19, opened 2024-07-25
Squad presets for all PoD warp locations

mayabyte:
Default squad and spray presets for PoD warp locations


jonathonbrady:
I'm kind of a fan of doing this the boring, verbose way and using a list of statically defined `Preset` structs that look like:
```c++
struct Preset {
    char* mMsgId;
    int mRedLeaves;
    int mRedBuds;
    int mRedFlowers;
    // ...
};
```
I think using compressed strings and writing custom parsing is cool but it sacrifices readability and extensibility to solve a problem we haven't demonstrated actually exists (insufficient on-console memory). I can see this getting unwieldy after adding Exploration Kit items, onion positions, Pikmin colors unlocked, etc.

mayabyte:
While there is a small memory advantage to doing it with PikiContainers, the actual motivations were 
a) to be able to apply the changes easily (e.g. it's just `playData->mPikiContainer = preset.mSquad` instead of having to add every pikmin type one at a time to a new PikiContainer with a loop for each one, for fifteen variables), 
b) to reuse PikiContainer's pre-existing serialization later on when we make these saveable, and 
c) to make _defining_ large amounts of presets easy and concise. 

The custom parsing (that last point) isn't strictly necessary here; we could do it with a bunch of like smaller builder pattern functions I guess? I do agree that the parsing function isn't exactly the nicest thing ever, it's just, if we ditch the string definitions then the preset definitions file is gonna turn into probably 200+ lines of samey builder stuff, which also kinda sucks I think. I can give it a shot tomorrow

jonathonbrady:
If you think builders would also suck then don't worry about changing anything - it works great and our time is better spent delivering value to the people etc. My thoughts are just about ease of development:
1. I think terse strings make defining large amounts of presets concise but not easy to read or write. We are going to add more things to presets which will require more parsing. It's not clear how we are going to represent, e.g., Exploration Kit treasures in strings in a vaguely readable way with parsing logic that doesn't suck to write.
2. I'm not convinced that a) or b) are sufficient reasons not to explicitly define struct members for each preset value. We can implement `Preset::getContainer` and apply the change just as easily with `playData->mPikiContainer = preset->getContainer()`. There's no avoiding adding 15 values one at a time to a `PikiContainer` - that's already being done by calling `Preset::_addByCharCode` for each Pikmin type anyway.
3. Hundreds of lines of builders does suck but it is extremely easy to find and edit the preset you are looking for and to make new ones without wondering if you wrote the string correctly.

jonathonbrady (approved):

PR #22, opened 2024-07-29
Retry cave feature + lots of polish

mayabyte:
- Press Dpad Down during the hole-in cutscene to retry the cave from the beginning (this resets the cave timer)
- Sublevel and cave retry works during the geyser cutscene now
- Prefix member variables with `m` (sorry HP)
- Extract warping functions from warp menu and reuse them for retry warps
- Record area and destination IDs in SegmentRecords
- Always initialize the cave timer when it hasn't been initialized (no more big silly number)


include/GlobalData.h
```diff
@@ -29,6 +29,8 @@ struct SegmentRecord {
 
 	u32 seed;
 	int floorIndex;
+	int areaIndex;
+	int destinationIndex;
 	Game::PikiContainer squad;
 	s64 startTime;
 	s64 endTime;
```
jonathonbrady, line 36: nit: still missing `m` prefix

src/plugProjectKandoU/singleGS_CaveGame.cpp
```diff
@@ -194,31 +198,57 @@ void CaveState::exec(SingleGameSection* game)
...
 
 			// Reset money
-			playData->mCavePokoCount -= p2gz->bugPokosCollectedSinceLoad;
-			playData->mCavePokoCount -= p2gz->treasurePokosCollectedSinceLoad;
+			playData->mCavePokoCount -= p2gz->mBugPokosCollectedSinceLoad;
+			playData->mCavePokoCount -= p2gz->mTreasurePokosCollectedSinceLoad;
+
+			// Reset squad to the one we entered the cave with when restarting from floor 1
+			// If we didn't enter from floor 1, just use our current squad.
+			// TBD: is this desired behavior?
```
jonathonbrady, line 234: We don't have to make a decision here but I don't think we should use the current squad since you will want to reset any deaths/growths. I think the viable options are:
1. Use the preset for sublevel 1 (annoying if you used a custom squad)
2. Use the squad you entered with (annoying if you started after a growth floor)
3. Reset to the sublevel you entered from with the squad you entered with (annoying if you want to restart from sublevel 1)

jonathonbrady (approved):

src/plugProjectKandoU/singleGS_CaveGame.cpp
mayabyte: I think this is kinda an edge case anyway so we can just like wait and respond to feedback about it when we eventually release the hack if we want. There is one more alternative though, and that's to make the retry cave option retry you from the sublevel you entered the cave from initially rather than always going to floor 1. If you're e.g. practicing GK and you want to skip GK1 every time that feels more sensible anyway? But yeah not a now thing

PR #24, opened 2024-08-02
Add upgrades to presets + more efficient flag repr

mayabyte:
Adds Exploration Kit upgrades to warping presets, plus uses a more space efficient data structure for storing lists of flags.


jonathonbrady:
Why reimplement `BitFlag`?

mayabyte:
ngl I didn't even think about looking at it before I went and wrote my own thing... 🤦‍♀️

mayabyte:
I tried it and it refuses to not crash no matter what I do, so I'm gonna call that a W for my gut feeling

PR #29, opened 2025-10-11
Add initial structure for GZ menu plus some silly demo things in it

mayabyte:
image
image

~~there are some invalid reads that I couldn't figure out so I'm gonna make someone else do it :)~~ I fixed them

double-press Dpad left to open or close the menu. should work anywhere I think


jonathonbrady (commented):
Very clean overall. Way better than the cowboy-oriented programming I would have done. Can you document where the invalid reads are (either in comments, or at least in the PR description) so we aren't confused why they're happening later?

PR #47, opened 2025-10-15
perma unlock ch mode + unlock all ch levels

amaritnez:
Challenge mode is available by default, and all levels are available to play on start.


HeartPiece44 (commented):
Nice, thanks for this! Minor request, can we keep any vanilla lines commented out, rather than deleted where possible? Just makes it so much easier to troubleshoot and temporarily revert things down the track.

PR #48, opened 2025-10-15
Enable treasure cutscene skipping

HeartPiece44:
Creates a menu item and associated managing class for toggling whether treasure collection cutscenes are skippable by pressing start. Handles both cave collection and above-ground collection, and avoids treasures double-counting when skipped after they hit the ship. Toggle is currently on by default.

Also makes pelletState.cpp equivalent in the process, which should help with pellet manipulation in the future.


mayabyte (changes requested):
Very nice. Just want to make sure we leave the original versions of those two lines as comments but then we're good to merge

HeartPiece44:
Very much agree - updated with comments of previous code!

mayabyte (approved):
SHIP IT

PR #49, opened 2025-10-15
Gate editor

mayabyte:
Can raise and lower segments of each gate individually, and adjust the current segment health with a FloatRangeOption (not visible in the video because I made it before adding that option). Gates are given names while above-ground based on a lookup table, and underground they're numbered.

https://github.com/user-attachments/assets/8eaac4fa-85a2-49eb-ad08-cce930afbdf3

### DIFF
--- a/src/plugProjectKandoU/itemGate.cpp
+++ b/src/plugProjectKandoU/itemGate.cpp
@@ -132,6 +132,12 @@ void ItemGate::onSetPosition()
 		efx::ArgRotY arg(mPosition, mFaceDir);
 		mEgateEfxBC->create(&arg);
 	}
+
+	// @P2GZ - Structure editor
+	// Register created gate with structure editor.
+	// Done in onSetPosition because StructureEditor uses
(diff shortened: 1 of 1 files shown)


configure.py
```diff
@@ -1305,7 +1305,7 @@
             Object(NonMatching, "plugProjectKandoU/gameMapParts.cpp"),
             Object(Matching, "plugProjectKandoU/vsGameSection.cpp"),
             Object(Matching, "plugProjectKandoU/gamePlatMgr.cpp"),
-            Object(NonMatching, "plugProjectKandoU/itemGate.cpp"),
+            Object(Equivalent, "plugProjectKandoU/itemGate.cpp"),
```
mayabyte: had to change this for my changes to take effect and I left it in the PR so others can pull the branch and try it out easily

mayabyte:
Fixed the crash by adding gates to the menu in `ItemGate::onSetPosition` instead of `ItemGate::doLoad` since the latter isn't called for gates in caves. Also removed the print

PR #56, opened 2025-10-20
add spray editor

jonathonbrady:
https://github.com/user-attachments/assets/2692fd9f-a25b-4117-84aa-d5b01e315073


amaritnez (changes requested):
2 Bugs:

Bug 1 - the spray setting in p2gz is desynced from the spray count. This is most noticeable if you use some sprays then warp between areas (See attached screenshot)

Bug 2 - in the overworld, the HUD for the sprays does not appear immediately when you unlock them. A video is attached to show this. I noticed after recording (not included in the video) that opening the map fixes this, but ideally we shouldn't expect people todo that in the first place :)
spray editor bug 1

https://github.com/user-attachments/assets/7345f103-9806-49b0-901b-1ac65060d4da

amaritnez (approved):
We can keep the small bug for now. I will track it in issues though just so if we ever have free time in the future we can fix it.

PR #57, opened 2025-10-21
Implement in-line image rendering in menus

HeartPiece44:
Adds image drawing within menu options, and associated image managing classes. Images are added to `assets/files/menu_images` as .btis, and added to the `ImageMgr::init()` function with an appropriate name string. Images are called from the manager using the name string and printed. A test menu option under Settings shows the image printing.

Currently, images are locked at a fixed size - functionality could/should be added to dynamically adjust that alongside the menu resizing option at some stage - open to whether that stage is now or later.

Image printing itself is by top-left coordinates, but is adjusted within menu option drawing to be bottom-left instead - this could be adjusted to be vertically centred on the line instead, or something else - open to ideas.


amaritnez (approved):
Seems stable enough. Played around with it for a bit and haven't noticed any new crashes. Memory is running tight but I feel this is a p2gz as a whole issue and not tied to this specific PR.

PR #65, opened 2025-10-21
Add toggle to skip save prompts in caves

HeartPiece44:
Does what it says on the tin - re-implements #25 (but without the crashing). Currently under Settings, and skipping save prompts is off by default - could change it so skipping is the default behaviour, but figured for now we'd want saving on for testing - lemme know if we should change this.

Probably scope for a future PR to update the timer to account for save prompt skipping.


mayabyte (changes requested):
Existing code looks good but it doesn't work for the first load after entering a cave from the surface. Probably just need some similar changes in `singleGS_MainGame`

HeartPiece44:
Good point, thanks for spotting that! Have implemented it in the same way and tested it.

Issue #63, opened 2025-10-21
Bug: warping into caves sometimes brings no Pikmin along


HeartPiece44:
Warping to caves has the chance to not bring your squad of Pikmin with you, causing an instant cave exit/extinction. Seems inconsistent, maybe a race condition?

PR #67, opened 2025-10-21
Fix warping to caves not bringing your squad with you & add an option to allow being in caves with no pikmin

mayabyte:
The old warp to cave code was far too restrictive about when it would bring your pikmin with you. It now properly brings all alive pikmin on the map with you into the cave when warping.

Additionally, this adds an option to allow being in caves with no pikmin and it's enabled by default. This reduces friction when just warping around trying to test features or look at things.

Fixes https://github.com/p2gz/p2gz/issues/63


amaritnez (changes requested):
Small bug: when you warp into a cave when you have wild pikmin, they will warp with you even though you haven't unlocked them (See video)

https://github.com/user-attachments/assets/7027c25f-ebc4-4dd3-9e3d-dbe175296364

amaritnez (changes requested):
I have been holding off on this but now that warping with wild pikmin is fixed and this bug still appears, it is worth noting: when you warp out of a cave with bulbmin, they will add to your pikmin total (the rightmost counter) even though they aren't in your party anymore. The video shows the following: entering FC3 with 20 reds. Recruit 10 bulbmin. Lost 2 pikmin, then exit back into the overworld. My total counter says I have 28 pikmin, despite only having 18 reds. This occurs even if you never whistled them (not shown in the video).

https://github.com/user-attachments/assets/2f160b33-ecc0-4849-82d5-dbd3e1095fe6


Additionally, warping when bulbmin are in the squad only sometimes removes them? You mentioned they are never carried forward, but the fact that it happens sometimes and is inconsistent is cause for concern.

PR #70, opened 2025-10-22
Input and display seeds for cave generation

mayabyte:
- Adds a seed input field in the warp menu
  - Uses a new `HexInputOption` that shows a 4x4 keyboard of hex digits
  - Allows reverting to a random seed
  - Use L and R to change which digit you're writing to, use the dpad to move on the keyboard, and press A to select a digit
- Displays the current sublevel seed in the pause menu and on the loading screen for the floor
  - (In the loading screen, it shows up on a delay because the map gen doesn't begin immediately)
  - Done in a new file `SegmentHistory`. This doesn't have any history-related code at the moment but it will be where it goes in the future
- Adds a `BoundDelegate` class to simplify certain menu callbacks

image
image
image

### DIFF
--- a/src/p2gz/gzmenu.cpp
+++ b/src/p2gz/gzmenu.cpp
@@ -535,6 +535,7 @@ void HexKeypad::set_unselected()
 
 void HexKeypad::submit()
 {
+	is_unselected = false;
 	p2gz->menu->pop_layer();
 }
 


amaritnez (approved):
I noticed a small bug that I will leave up to you if you decide is breaking or not (this feels mostly QoL so I'm not blocking the PR for it): if you select a seed, then go to random, then go back to set a new seed, the new seed will not save until you input a number. In other words, if you make a test seed (lets say 0x12345678), then go to random, then want to go back to 0x12345678, you need to re-input at least one number before clicking submit, or else the menu will still set it to random.

Otherwise, everything seems to be working fine.

Issue #60, opened 2025-10-21
Bug: collision viewer crashes in caves and VS mode


HeartPiece44:
Game crashes when toggling on the collision viewer while in caves, or while in a different game mode such as VS mode. Could be related to the polymorphic mapMgr?

PR #72, opened 2025-10-22
fix: handle mapMgr polymorphism in CollisionViewer

jonathonbrady:
`mapMgr` is an instance of `ShapeMapMgr` in the overworld, but `RoomMapMgr` in caves, so the static cast to `ShapeMapMgr` crashed the game when enabling collision viewer in caves.

Closes https://github.com/p2gz/p2gz/issues/60.


amaritnez (changes requested):
As discussed, collision viewer overlaps in floors with a solid plane (sh2, gk1, etc.).
image
image

Try looking at the `mFloorHidden ` field in FloorInfo to see if that could help get you anywhere.

HeartPiece44 (approved):
Nice! Collision seems to hide nicely in places like GK2 now. Did get a random crash going OoB to the VRBox collision with the viewer on, but couldn't replicate it, so maybe we just keep an eye on it down the line. Also, the game chuuuuuugs when walking over certain collision or doing certain things (such as the railroad tracks in GK2, or when using "boing") with the viewer on, but that's also something we can look into later.

PR #82, opened 2025-10-26
Add an issue template

mayabyte:
so issues don't stink


amaritnez (commented):
This im more hesistant on: is there a reason why we wanna make a file for tracking bugs instead of using github's issue tracker?

PR #83, opened 2025-10-26
Add a changelog with alpha1 features

mayabyte:
We should update this with features and bug fixes every time we make a new release.


amaritnez (approved):
Makes sense to me, lets add it!

PR #85, opened 2025-10-28
Fix: Make Pikmin join squad organically with squad editor spawning

HeartPiece44:
Currently, Pikmin spawned with the squad editor spawn in a weird AI state, and only join the squad on bump/with some delay. This normalises the behaviour so they integrate into the squad as expected.


jonathonbrady (approved):
👉🚢

PR #86, opened 2025-10-28
p2gz title screen logo

amaritnez:
Updates the "Pikmin 2" text on the title screen to say "P2GZ", complete with updated colors.
image

Note that cube currently fails to pack the "title.szs" file correctly. Doing so with cube causes the colors to be corrupted; I used the not-cube compression tool to make the file, and then am having the builder just copy the .szs file directly vs. compressing the contents. We should fix this at some point.


amaritnez:
image


https://github.com/user-attachments/assets/46330881-cfdc-4e51-bfab-d4b9b8310754

Updated title

PR #89, opened 2025-10-29
Bug fix: remove crash on re-entering area

### LINKED ISSUE #71: Crash when you enter an area for the 2nd time
If you enter an area, end day, save, then renter an area the next day you crash.

https://github.com/user-attachments/assets/6fd8b3b9-9817-4456-b1c3-628f25207c98

HeartPiece44:
Fixes #71 by adjusting how the Vec implementation clears its members, especially those with non-trivial destructors, and forcing dynamic menu clearing on load to use the Vec method to avoid an infinite loop.

### DIFF
--- a/include/p2gz/StructureEditor.h
+++ b/include/p2gz/StructureEditor.h
@@ -39,7 +39,7 @@ public:
 private:
 	const char* get_gate_name(f32 x, f32 z);
 
-	Vec<GateWrapper> gates;
+	Vec<GateWrapper*> gates;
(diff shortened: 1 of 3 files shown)


HeartPiece44:
Good spot, thank you - I think it seems to work now for both warping to areas and sublevels, as well as re-entering the same area. Got fooled into thinking it was a Vec issue when I think it was actually a local variable copy issue.

mayabyte (approved):

PR #94, opened 2025-10-29
Script for building release zips

mayabyte:
Adds a script `release.py` that performs a clean build of the hack and then copies everything necessary to install the hack into a `p2gz-VERSION.zip`. Based heavily off the installer used in VS Max. 
Automatically reads the version from `include/p2gz/p2gz.h` so we only have to set it in one place.


amaritnez (changes requested):
Currently the script just duplicates the vanilla copy of P2 vs. actually patching and making a p2gz iso

PR #97, opened 2025-10-30
Improve warping preset logic

mayabyte:
Presets are now suggested and used in a more sensible way. Deliberately chosen presets take priority over suggested and generated ones, but after being used all presets are marked as "stale" and other presets will override them. This feels a bit more natural to use.

Also fixes a bug where the wrong preset was being searched for for SCx 1-4. Adds a log so it's easier to catch instances like this in the future.


amaritnez (approved):
Seems fine to me. Small bug but I believe this is on main, but it is warp related:
If you warp to a final floor of a cave, then take the geyser and reset cave on the final floor (press L), then use warp to go back to the final floor, it doesn't take you there but instead to sublevel one. It saves your warp destination but doesn't actually apply. You need to manually readjust the sublevel back to the final floor for it to take effect.

PR #99, opened 2025-10-30
Add EK upgrades to presets

mayabyte:
Adds EK upgrade field to presets and populates appropriate values for all PoD presets


amaritnez (approved):
Seems good to me, no real issues. GG

Issue #91, opened 2025-10-29
Double press and Press-and-hold actions are twice as fast on the menu


mayabyte:
**Description of the bug**
Since the menu runs at 60FPS, all gzmenu actions that use frame-based timing are twice the speed.

PR #100, opened 2025-10-31
Support warping from file select

mayabyte:
Allow warping from file select just like any other menu.

Also fixes https://github.com/p2gz/p2gz/issues/91

### DIFF
--- a/include/p2gz/InputHelpers.h
+++ b/include/p2gz/InputHelpers.h
@@ -3,6 +3,7 @@
 
 #include <Controller.h>
 #include <System.h>
+#include <p2gz/HelperInlines.h>
 
 namespace gz {
 
@@ -17,8 +18,6 @@ public:
 
 	bool check(Controller* controller)
 	{
-		const bool is_30_fps = sys->mFrameRate == 2.0f;
-
 		if (frames_left > 0) {
 			frames_left -= 1;
 		}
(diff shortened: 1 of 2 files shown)


mayabyte:
Now ready for review

amaritnez (changes requested):
Save between caves is gone now. Save prompt is still present when exiting a cave, but should we just tell people not to save for now, or would you like to remove it?

Additionally, since it is a "fresh" save, many cutscenes will still play (such as first treasure, first cave enter, etc.). This is worth blocking as it can make practicing caves impractical, especially since there is no saving so this occurs all the time. The most agregious example is pikmin are now immune to being eaten because the flag for disabling their immunity is not set yet.

include/p2gz/InputHelpers.h
```diff
@@ -16,6 +17,8 @@ class DoublePress {
 
 	bool check(Controller* controller)
 	{
+		const bool is_30_fps = sys->mFrameRate == 2.0f;
```
jonathonbrady: nit: Since you already have to rebase, it might be nice to add this to the helper inlines header.

jonathonbrady (approved):

amaritnez (approved):
approve for alpha1 (prayage for no crash).

Remind people not to save in bad spots or else

PR #103, opened 2025-11-01
Overhaul timer functionality

### LINKED ISSUE #61: Enhancement: improve timer formatting
Timer displays as 2:30:5 for 2 min 30 s 500 ms, and should probably display as 2:30.5 or something similar.

HeartPiece44:
Overhauls timer (and closes #61) by implementing:
- better timer formatting (decimal point for tenths and less horizontal jostling)
- improved reset logic based on game state and logical split times
- displaying and hiding subtimer automatically when in/out of caves
- handling save prompt skip timing offsets
- handling treasure cutscene skip timing offsets (dynamically based on cutscene skip time)
- pausing timer when gz menu is open

Also adds some state-checking helper functions for future use.

### DIFF
--- a/include/p2gz/HelperInlines.h
+++ b/include/p2gz/HelperInlines.h
@@ -7,37 +7,37 @@
 
 namespace gz {
 
-inline bool inBootUp()
+inline bool in_boot_up()
 {
 	// make sure gameflow is in boot up state
 	return GameFlow::mActiveSectionFlag == GameFlow::SN_Boot;
 }
 
-inline bool inTitleScreen()
+inline bool in_title_screen()
 {
 	// make sure gameflow is title state (incl. options menus, hiscore menus, etc)
 	return GameFlow::mActiveSectionFlag == GameFlow::SN_MainTitle;
 }
 
-inline bool inStoryMode()
+inline bool in_story_mode()
 {
 	// make sure game system is active and we're not in challenge mode/vs mode/piklopedia/etc
 	return Game::gameSystem && Game::gameSystem->isStoryMode();
 }
 
-inline Game::SingleGameSection* getSGS()
+inline Game::SingleGameSection* get_SGS()
 {
 	// make sure current section is castable to SingleGameSection
-	if (!inStoryMode()) {
+	if (!in_story_mode()) {
 		return nullptr;
 	}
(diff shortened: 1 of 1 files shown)


src/p2gz/timer.cpp
```diff
@@ -56,34 +73,188 @@ Timer::TimeComponents Timer::calc_time(u32 start_time)
...
+	reset_sub_timer(offset_seconds);
+}
+
+void Timer::offset_main_timer(f32 offset_seconds)
+{
+	if (offset_seconds >= 0.0f) {
+		// "add" to timer
+		u32 secs      = floor(offset_seconds);
+		u32 tenths    = floor((offset_seconds - floor(offset_seconds)) * 10.0f);
+		u32 millisecs = secs * 1000 + tenths * 100;
```
mayabyte: These two calcs are the same right? could we do them once outside and just flip the sign when adding to `main_timer`?

HeartPiece44:
Have also implemented logic for new file selection, both with and without memory card. The timer starts nicely on A press on selecting a fresh file with memory card, but it's the smallest bit delayed for no memory card. Also, I've had to force the timer to draw over the crash landing cutscene, since currently it doesn't - this does result in the timer getting stolen by the talking pod during the text, however, but unsure how to fix - maybe better than no timer during the cutscene? Help appreciate lol:

image

HeartPiece44:
Fixed hostage situation between timer and talking pod, prints as expected now.

amaritnez (changes requested):
Major bugs, I feel this are worth blocking the PR over:

- Timer doesn't reset on new day (at least sub timer should)
- Timer doesn't pause when using freecam (local teleport ability)
- Timer resets if you choose to hide it (enable/disable option - ideally it just hides the timer vs. resetting the value outright)

Minor bugs, feel free to ignore if you want for now but eventually should be fixed:

- Timer doesn't play during louie skip cutscene
image

- sub timer goes away when exiting a cave via geyser (is still counting, but needs to be reshown)
image

mayabyte (commented):
I haven't gotten around to testing this yet but it's looking good, only some small comments on the code

jonathonbrady (approved):

amaritnez (approved):
Approve for alpha1 (should hopefully work)

PR #108, opened 2025-11-01
Respawn treasures in caves on any warp, not just cave warps

mayabyte:


amaritnez (approved):
Works now, fixes the issue described, and your implementation seems solid. ship

PR #109, opened 2025-11-01
Put cutscene flags in presets

mayabyte:
yop


amaritnez (approved):
Approve to prep for alpha1

PR #113, opened 2025-11-02
Force p2gz allocations to happen in the system heap so they don't get clobbered

mayabyte:
make sure to use the system heap for p2gz allocations so our memory doesn't get clobbered

use `last_used_preset` to set cutscene flags etc. when warping. when there is none, use values from recommended preset


HeartPiece44 (approved):
Thank you for fixing this stuff up! Seems to work as intended now - cutscenes seem to hold, retrying sublevels doesn't break anything, and warping to above grounds seems to destroy gates as intended. Would be good to get some more testing on it from others before merging though imo, given the Horrors, but looks good so far c:

amaritnez:
Update on the 1hp: it appears this only happens if you warp to a cave on file select. Warping to an overworld has the navi hp work like normal. Also here's a screenshot of how it looks:
image

PR #116, opened 2025-11-03
dismiss circles now wrap around the ground

amaritnez:
"Dismiss positions" option now has the circles wrap along the ground. Note that circle positions can get a little iffy with dramatic slopes/walls/oob areas, want to get everyone's take on it. Playtesters (jpep) complained about the original circles being flat and clipping into the ground, so this helps to address that. Here are some images with the new circles (both good and bad).

image
image
image
image
image
image
image
image
image
image
image


mayabyte (approved):
Works great now. There's still a little clipping in some spots but that's fine - the visibility issues are gone.

Issue #120, opened 2025-11-03
No triangles drawn while swapping captains with collision viewer open


HeartPiece44:
**Description of the bug**
Swapping captains with collision viewer open causes no triangles to be drawn during the swap, which is disconcerting. Ideally, we'd connect the position of the collision view render sphere to the camera or something instead of the navi, since the navi position won't be as expected while swapping.

**Steps to Reproduce**
1. Turn on collision viewer
2. Swap captains back and forth

**Crash Log or Screenshots**
https://www.twitch.tv/jpeppersons/clip/RichTransparentYamLitty-fYcC2G_-FwodDMEN

PR #140, opened 2025-11-03
fix: render collision while switching captains

jonathonbrady:
Bonus side effect of highlighting either Navi's floor triangle, instead of just the active Navi.

Closes https://github.com/p2gz/p2gz/issues/120.


HeartPiece44 (approved):
Looks good! Does what it says on the tin, nice - seems to work well.

Issue #75, opened 2025-10-23
Opening the GZ menu while in a loading screen leaves the game unpaused when the load finishes


mayabyte:
Steps to reproduce:
- Enter a loading screen e.g. by going to the next floor in a cave
- Open the GZ menu before the next floor loads

PR #144, opened 2025-11-04
fix: limit when you can enter p2gz to prevent unpausing desync issues

amaritnez:
Restricts when p2gz menu can be opened. The following no longer allows for p2gz menu:
- cutscenes
- loading screens
- end of day
- day results
- cave results

This should help with issue #75 , as the game won't unpause if it never pauses to begin with.

### DIFF
--- a/src/p2gz/gzmenu.cpp
+++ b/src/p2gz/gzmenu.cpp
@@ -256,7 +256,12 @@ void GZMenu::close()
 		return;
 
 	enabled = false;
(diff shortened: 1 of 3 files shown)


mayabyte:
I think a sensible fix for the above would just be to force-close the menu when entering the holein scene

HeartPiece44 (approved):
Nice work! Soothes a lot of anxiety going into more play tests - very happy people can still warp from cave results and day end results, super smooth.

Issue #126, opened 2025-11-03
Using Captain->kill option unpauses gameplay while menu is still active


HeartPiece44:
**Description of the bug**
Using the Captain->kill menu option causes the gz menu to remain open while unpausing gameplay, making it easy to accidentally kill both captains - should either close the menu on selecting the option or re-pause gameplay after captain death screen.

**Steps to Reproduce**
1. Open GZ menu
2. Select captain->kill option
3. Game is now unpaused while menu is open

PR #150, opened 2025-11-04
Captain menu improvements

mayabyte:
- Changes the boing option to "boing mode". Pressing A while it's enabled gives you a small but more manageable vertical boost.
- Menu closes when selecting the kill option. Closes https://github.com/p2gz/p2gz/issues/126
- Adds a health editor for the current captain


amaritnez (changes requested):
2 issues, one crash and one slightly minorr:

Crash: If you kill a captian, then warp somewhere, notice how both captains are revived. If you then kill the other captain, the game will think you killed 2 captains even though you didn't (either end day or exit cave prematurely). Then once you exit the respective menu, the game crashes because it's confused on whether the captains are alive or not. See video for example:

https://github.com/user-attachments/assets/5734f465-f887-4249-8ed1-086b80d9bf7d

Minor issue: If you set a captain's HP to 0 via holding X, it does the same weird glitch where the game "thinks" they are dead but in reality are not. Ideally if you set the HP to 0, the captain should just die rather than being in this glitched state (see image):
image


Also small thing: imo, boing should be a hold A to fly instead of a button press, but open to suggestions from others.

HeartPiece44 (approved):
Looks good! The previous bugs all seem fixed. Boing mode is very fun - I think I agree that in the future it'd be good to improve the air movement, so you're not just stuck in one direction, and so you don't lose all your momentum when you hit a wall, since that makes jumping over invisible vertical walls a bit annoying - but agreed, low priority + for future.

Issue #121, opened 2025-11-03
Waypoint viewer position moves slightly with camera


HeartPiece44:
**Description of the bug**
When moving the camera around with the waypoint viewer on, the positions of the cones and lines moves slightly, which is disconcerting - should probably be fixed relative to map model.

**Steps to Reproduce**
1. Warp to AW AG
2. Turn on waypoint viewer
3. Walk around and toggle camera angles to see waypoints move slightly

PR #152, opened 2025-11-04
Cave spawnpoint viewer

mayabyte:
Adds a spawnpoint viewer for caves. Uses the same colors as caveripper. Currently the circles don't wrap to the floor, so it'll be worth updating this code after https://github.com/p2gz/p2gz/pull/116 is merged.

Also fixes https://github.com/p2gz/p2gz/issues/121

image


amaritnez (commented):
Everything seems sound. Will await the merge from PR #116 as without it some circles float in the air and are very awkward (see image). Otherwise, no issues at the moment.
image

amaritnez (approved):
Seems like everything working and the code seems fine. Some small issues/suggestions, but nothing to block over.

First: perhaps we should make some of the circles wider? For larger items they cover the circle making it harder to see the center (see image)
image

There is a small graphical glitch where the spawnpoints like disappear for a sec if you put the camera high (see video). Nothing big, but just thought I'd point it out

https://github.com/user-attachments/assets/6343b0f2-90b8-4ed7-85c5-e7252f90412e

Overall good job though.

PR #153, opened 2025-11-05
partially revert #147

jonathonbrady:
Reverts the preset menu auto-scroll and nested delegates added in https://github.com/p2gz/p2gz/pull/147.


amaritnez (approved):
No longer crashes, ship

Issue #148, opened 2025-11-04
Adjusting time can trigger the "first sunset" cutscene and unpause the game while the GZ menu is open


mayabyte:
**Steps to Reproduce**
1. Open gz menu while AG
2. Adjust time of day to near the end
3. "first sunset" cutscene should play
4. Game will now be unpaused while GZ menu is open

PR #154, opened 2025-11-05
fix: closes timer menu when sunset cutscene plays

amaritnez:
Fixes issue #148 by making it so that the p2gz menu will force close once the sunset cutscene plays. Currently awaiting on PR #144 to be approved, as currently you can enter the p2gz menu while the cutscene actually plays (PR #144 fixes that). For now, this will remain a draft PR until the other cutscene fixes are implemented.

### DIFF
--- a/src/p2gz/timer.cpp
+++ b/src/p2gz/timer.cpp
@@ -291,7 +291,7 @@ void Timer::set_sub_timer_enabled(bool on)
 	sub_timer_enabled = on;
 }
 
-// Yeehaw the follow code from hurryUp2D.cpp since it's not matching and we need it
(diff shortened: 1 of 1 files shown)


amaritnez:
Ok, it should be ready now that it has the features from PR #144 .

src/p2gz/timer.cpp
```diff
@@ -281,3 +290,48 @@ void Timer::set_sub_timer_enabled(bool on)
 {
 	sub_timer_enabled = on;
 }
+
+// Yeehaw the follow code from hurryUp2D.cpp since it's not matching and we need it
```
HeartPiece44: can we change this into a `// @Extracted: hurryUp2D.s scaleUp2__Q28Morimura10THurryUp2DFv` so we can trace where it came from a little easier?

PR #155, opened 2025-11-05
Implement bridge, plug and bag editors + bridge and plug debug info

HeartPiece44:
Adds bridge, plug, and bag editors to the structure editor + menu, along with debug info toggles for both the bridge and plug editors - doesn't seem to be much point in one for the bags. 

Doesn't add presets for these yet, but should be fairly straightforward to do based on the gate presets. 

Plugs are currently a bit iffy - they usually are killed by the manager rather than just hidden, so the models get a bit weird. I've also had to implement some logic to re-fill the water boxes around the plugs, seems to work fairly nicely, but there's a slight delay. I've made it so the menu closes automatically on updating plug status, so people can't spawn 10 before they resolve, but I'm not entirely convinced it won't crash if someone really wants it to. I think that's a later problem to resolve, having the functionality seems more important at the moment, but fixes welcome - otherwise, I'll open an issue for it once this is merged. 

Also, plugs don't automatically update their associated hole effects or toggles yet either, so technically you could kill the SR plug, respawn it, and then enter the cave with the plug still alive. Again, bug issue for later, will open issue on merge.

### DIFF
--- a/src/p2gz/structureEditor.cpp
+++ b/src/p2gz/structureEditor.cpp
@@ -420,10 +420,12 @@ void StructureEditor::PlugWrapper::set_plug_state(bool alive)
 			if (strcmp(map.name, name) != 0) {
 				continue;
 			}
-			plug->mWaterbox->mBounds.mMin   = map.min_bounds;
-			plug->mWaterbox->mBounds.mMax   = map.max_bounds;
-			plug->mWaterbox->mLoweredAmount = 0.0f;
-			plug->mWaterbox->mState         = Game::AABBWaterBox::WaterBox_Active;
+			if (plug->mWaterbox) {
+				plug->mWaterbox->mBounds.mMin   = map.min_bounds;
+				plug->mWaterbox->mBounds.mMax   = map.max_bounds;
+				plug->mWaterbox->mLoweredAmount = 0.0f;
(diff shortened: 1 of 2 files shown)


src/p2gz/structureEditor.cpp
```diff
@@ -129,30 +204,421 @@ void StructureEditor::sync_gates()
...
+	for (size_t i = 0; i < NUM_BRIDGE_NAMES; i++) {
+		NameCoordinateMap map = BRIDGE_COORD_TO_NAME[i];
+		if ((absF(map.x - x) < STRUCT_SEARCH_RADIUS) && (absF(map.z - z) < STRUCT_SEARCH_RADIUS)) {
+			return map.name;
+		}
+	}
+
+	char* name = new char[8];
+	sprintf(name, "bridge %d", bridges.len());
+	return name;
```
mayabyte: There are never any bridges in caves so I'd suggest a `GZASSERTLINE` here instead so we notice issues more quickly if they crop up

amaritnez:
Additionally, burrownits seem to always have bridge glitch enabled? But I'm unsure if this is a vanilla bug or not. They are coded to only do 100 damage to bridges, but instantly delete them. 

https://github.com/user-attachments/assets/71f4359d-0375-413e-9d28-76faef1d087e

mayabyte (approved):
🚢 
☝️

PR #162, opened 2025-11-06
Enemy collision viewer

amaritnez:
Adds a new option under "enemy debug info" to draw collision spheres. Collision spheres are color coded to represent the following:

- Green: collision spheres that pikmin can latch onto. These are normally enemy weakpoints, or generally areas where pikmin can deal the most damage (bulborb butt, wollywog body, beady long legs head, etc.). Internally these are collision spheres with special attribute 'st__' (or in the case of bulborbs, 's___').
- Yellow: collision spheres that pikmin can target, but cannot latch onto. These vary by enemy. For some enemies, these represent areas where pikmin can attack but deal reduced damage (such as pileated's foot). For other enemies, this represents armored areas where pikmin will do no damage (such as flint beetle's body, burrow-nit's shell, beady long leg's feet, etc.). Internally these are collision spheres with special attribute '\_t__'
- Red: collision spheres that pikmin generally will not target and ignore. This includes intangible things like bulborb feet, honeywisp's body, waterwraith (and his rollers), and rocks. Exception: hazzards, eggs, and other simple "enemies" will use this collision type but can still be targetted. Internally these are collision spheres with special attribute '____'

image
image
image
image
image
image
image
image

### DIFF
--- a/src/p2gz/enemyDebugInfo.cpp
+++ b/src/p2gz/enemyDebugInfo.cpp
@@ -265,22 +265,7 @@ void EnemyDebugInfo::draw_enemy_dbg(Game::EnemyBase* enemy, Graphics* gfx)
 	}
 }
 
-/**
- * Uses both recursive and non-recursive iteration to draw all collision spheres in the colltree of an enemy
- * Collision tree in an enemy can be visualized as follows:
- * root
- * 		A
- * 		B
- * 			C
- * 			D
- * 		E
- * 			F
- * 				G
- * 		H
- * end
- *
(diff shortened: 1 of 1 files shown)


mayabyte (approved):
Works great

HeartPiece44 (approved):
Nice! Looks sick, super cool feature to have. 🚢 

Will note that eventually we need to sort out the prim draw stuff so that it doesn't move so much on camera move - it's especially pronounced for this, which makes things a bit difficult, but that's definitely out of scope to fix in this PR.

PR #164, opened 2025-11-06
Basic skeleton for integration tests

mayabyte:
Adds an extremely basic and feature-incomplete framework for integration tests. Tests take the form of a list of TestOps to be performed. TestOps can be single button presses, waiting for an arbitrary state to be reached, or in the future directly editing the game or menu state. There are some simple combinator TestOps to do more complex actions. Macros are provided to make writing tests (sequences of TestOps) simpler.

Test code is all behind conditional compilation `#ifdef`s. Pass the `--test` flag when building (e.g. `python3 build.py --test`) to build with testing code enabled. Building without the test flag should have zero (or extremely minimal) effect on binary size. Tests will begin running as soon as GZ is inited, which is currently upon entering file select. Sequencing info is printed to the log via OSReports. 

This is meant to be an _extremely basic first pass_ at how the integration tests should look. When we want to write feature tests, we will inevitably need to add more kinds of TestOps and more macros. I want to put this out there and possibly merge it in this early state to gather feedback and let everyone contribute to how this system looks and works as we write tests.

**I want three reviews on this PR before merging** to ensure we're all on the same page about this system.


include/p2gz/Test.h
```diff
@@ -0,0 +1,144 @@
...
+namespace test {
+
+// Creates a Test object with the right array sizes
+#define TEST(name, ...) new Test(name, ARRAY_SIZE(((TestOp*[]) { __VA_ARGS__ })), ((TestOp*[]) { __VA_ARGS__ }))
+
+#define PRESS(btn)     new ButtonInput(btn), new Wait(1)
+#define DBL_DPAD_L     PRESS(PAD_BUTTON_LEFT), PRESS(PAD_BUTTON_LEFT)
+#define DO_N(num, ...) new DoN(new CompoundOp(ARRAY_SIZE(((TestOp*[]) { __VA_ARGS__ })), ((TestOp*[]) { __VA_ARGS__ })), num)
+
+#define WAIT_FOR(state_fn) new WaitForState(new FreeDelegateR<bool>(&state_fn))
```
jonathonbrady: nit: Would be nice to have macros for even the trivial test ops so we don't have macros interspersed with `new Wait(300)`, etc.

mayabyte:
After my findings around binary footprint in https://github.com/p2gz/p2gz/pull/168 I'm going to take another pass at this before merging. The `TEST` macro in particular is likely very bad for binary size as-is.

PR #165, opened 2025-11-06
add treasure editor

jonathonbrady:
Stands up TreasureEditor. Currently supports moving treasures to arbitrary locations with freecam. Handles buried treasures, treasures held in enemies, captured by breadbugs, and carried by dweevils.

Toggling whether a treasure is collected or not currently does nothing. This is a separate effort and will be added in a future PR.

The code for treasure editor is currently duct-taped on top of freecam. I do not think this is a good implementation. We will probably use freecam in more places eventually, so in the future I want to refactor freecam's enable function to take a configuration with Delegates that run when you press buttons, but that is also a separate effort.

### DIFF
--- a/src/p2gz/gzmenu.cpp
+++ b/src/p2gz/gzmenu.cpp
@@ -408,12 +408,32 @@ void ListMenu::update()
 		selected = scroll;
 	}
 
+	u32 btn = p2gz->controller->getButtonDown();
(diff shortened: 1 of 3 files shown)


mayabyte (changes requested):
Teleporting with freecam leaves the game in a paused state. It can be fixed by opening and closing the gz menu but it's pretty disruptive

Issue #130, opened 2025-11-03
Retrying sublevel with bulbmin causes preset menu rendering issue


HeartPiece44:
**Description of the bug**
Having bulbmin either in squad or present on a sublevel, then retrying the sublevel, causes the preset menu to display an extra blue onion with their count, regardless of if they're in your squad or not.

**Steps to Reproduce**
1. Warp to FC3
2. Retry sublevel using hole or geyser
3. Open warp menu and see preset is now showing extra blue onion with bulbmin count (10).

**Crash Log or Screenshots**

<img width="1380" height="1008" alt="Image" src="https://github.com/user-attachments/assets/22545b0f-b9b2-423b-a0b2-348d47b59b14" />

PR #166, opened 2025-11-06
fix: wild bulbmin no longer show up in generated presets

mayabyte:
Closes https://github.com/p2gz/p2gz/issues/130

Also adds bulbmin images to the preset rendering code so they'll show up properly when they're actually supposed to be there.


amaritnez (approved):
Seems like it works, gg

Issue #132, opened 2025-11-03
Adjusting poko count in caves does not function intuitively


HeartPiece44:
**Description of the bug**
Adjusting poko count using the menu while in a cave causes the poko count display to appear, showing the incorrect amount - it will display the poko amount set with the menu + whatever pokos have been collected during the cave. It should either adjust the base amount so that base + current cave = amount set, or there should be a second poko editing option while in caves to adjust in-cave collection so far.

**Steps to Reproduce**
1. Warp to EC1, collect treasures or enemies
2. Open GZ menu and adjust pokos

Issue #107, opened 2025-11-01
Retrying sublevel doesn't reset poko count


amaritnez:
**Description of the bug**
Retrying a sublevel by using the X/Y feature when entering a cave doesn't reset poko amounts, meaning when you collect a treasure again it just adds to what you collected.

**Steps to Reproduce**
1. Collect a treasure in any cave (such as citrus lump in ec1).
2.  Enter the cave to go to ec2, but press retry sublevel (either same seed or random seed, does not matter)
3.  Collect citrus lump again (or anything, does not matter).
4. Notice the pokos still includes the first treasure you collected in addition to the newly collected item.

https://github.com/user-attachments/assets/c760f34a-bec6-4c0e-8d56-909adcdd7bd9

PR #167, opened 2025-11-07
Restore pokos when retrying sublevel + several poko editor fixes

mayabyte:
- Restore pokos when retrying sublevel. Closes https://github.com/p2gz/p2gz/issues/107
- Adjusting poko count in caves now works as expected. Closes https://github.com/p2gz/p2gz/issues/132
- Refactor PokoEditor to work via delegates instead of the update loop. DecimalInputOption now takes delegate arguments like HexInputOption.
- Add pokos and enter kind fields to Presets. (I haven't added poko values to the PoD presets in this PR).
- Fix a crash when selecting a preset category menu that doesn't match your current preset (I can't believe we didn't catch this)

### DIFF
--- a/src/p2gz/pokoEditor.cpp
+++ b/src/p2gz/pokoEditor.cpp
@@ -18,7 +18,7 @@ void PokoEditor::sync()
 
 void PokoEditor::apply_cave_pokos()
 {
-	Game::playData->mPokoCount     = get_pokos();
+	Game::playData->mPokoCount     = Game::playData->mPokoCount + Game::playData->mCavePokoCount;
 	Game::playData->mCavePokoCount = 0;
 	sync();
 }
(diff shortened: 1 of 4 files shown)


mayabyte:
Bugs are now fixed.

> Also, not a bug but moreso a wider discussion/question: now that pokos can be edited in caves, if you edit pokos in a sublevel and then retry, it will overwrite your poko count to what you had before you retried. I'm wondering which should take priority here? I'm indifferent either way, but it is something we should keep in mind.

It should absolutely reset to the pokos you had before retrying. Otherwise if you retried a sublevel a few times and then continued on into later splits, you'd have an incorrect poko count.

> Also also, small thing but it would be nice if you could make it so the poko editor wraps back to the most significant digit once you edit the final number (no big deal if you don't do this now).

Can't be bothered right now but just to make sure, you know you can use L and R to move the cursor right?

mayabyte:
@amaritnez this bug is the combination of another known bug (restarting the cave sometimes doesn't use the generated preset and falls back to the suggested one) and the lack of poko values in our current pod presets. Solving both of those things would resolve this, so I think it's not an issue that needs to be addressed in this PR specifically.

HeartPiece44 (approved):
Looks good to me! Appreciate all the bug fixes alongside this 🚢

PR #168, opened 2025-11-08
Respawn treasures and enemies AG

mayabyte:
- Treasures and enemies can now be respawned via presets
- PoD presets have been adjusted with mostly correct enemy and treasure spawns. There are some small discrepancies (no way to specify a position for Ursidae on enter FC for example) but those can be addressed in a future PR as this one is already way too big.
- Adds basic generator debug info similar to enemy debug info. Currently only shows name and position but can be expanded in the future.
- Fixes an issue with treasure moving where they'd be placed halfway in the ground.
- Sync treasure menu upon opening
- Replace menu images for pikmin colors with 32x32 variants
- Add destructors for all gzmenu types
- Optimize binary size for `presetMgr.cpp`
- Adds debug memory prints that we should leave in to keep track of our memory footprint


mayabyte:
> Structures and enemies don't seem to respawn properly. 

This is happening on main as well apparently. This branch doesn't have the structure editor PR in so it was around before that and none of us noticed somehow.

mayabyte:
@amaritnez The above bugs have now been fixed by using a completely different method to respawn treasures.

amaritnez (dismissed):
Respawning treasures doesn't seem to work now? Note sure waht are the conditions to get them to respawn. I collect the spiny alien treat, end day, then warp to the same day again and realize the treasure hasn't respawned yet.

https://github.com/user-attachments/assets/f8d6416a-91b3-4ccb-bcbb-e38eef5e116f

PR #171, opened 2025-11-09
Add presets for all structures

mayabyte:
- PoD presets for all structures
- Fixes an issue where structures that were previously destroyed wouldn't respawn


amaritnez (changes requested):
Very small thing: the enter FC splits dont take into account the plug. Otherwise everything seems fine so far

amaritnez (approved):
Works now, gg

PR #173, opened 2025-11-09
fix: boing mode no longer stores upward velocity from A presses in gz menu

mayabyte:


amaritnez (approved):
Works.

2 boing bugs but not tied to this pr we should fix: 1-boing still stores A presses from navigating cutscene menus

2- boing crashes if you switch captains and press A during the transition animation (active_navi() does not work when switch captains)

PR #176, opened 2025-11-09
Add dynamic control display for menus

mayabyte:
Also slightly adjusts some of the menu colors to have better contrast and be more pleasant to look at.

image
image


src/p2gz/gzmenu.cpp
```diff
@@ -24,6 +24,10 @@
 
 using namespace gz;
 
+#define COLOR(color)           \
```
jonathonbrady: nit: can we call this `SET_COLOR` or something?

HeartPiece44 (approved):
Definitely cooked with this one, looks great - appreciate all the work! I also think we might need to adjust the yellow for squad selection down the track, it's pretty low-contrast on, say, the grass area of AW vs the white, so it's easy to lose where your cursor is - but we can tweak that down the line.

PR #186, opened 2025-11-12
Add day numbers to presets

HeartPiece44:
All current presets now have associated day numbers which auto-update the warp menu selection. Selecting a new day number from the warp menu after setting a preset will override the preset number (and vice versa). Also splits the EC preset into EC1 and EC2 due to cutscene flags.

(Also adds the framework for presets for Days 1-4 (mainly cutscene flags), but the implementation is currently commented out since we don't support sprouts yet, and we may have to do some shenanigans for day 1 new file/day 2 extinction load-in. Hopefully it just saves someone some effort down the line.)

### DIFF
--- a/include/p2gz/warp.h
+++ b/include/p2gz/warp.h
@@ -87,6 +87,7 @@ private:
 	void update_cave_opt();
 	void update_sublevel_opt();
 	void update_preset_opt();
+	void update_day_opt();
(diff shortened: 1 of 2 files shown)


mayabyte:
Hiding the day option unless you have no preset selected seems like the right approach for now, yeah. Should be a simple change.

PR #188, opened 2025-11-12
Support using sublevel retry in gz menu

mayabyte:
Supports using sublevel retry and retry same seed options while in a cave in the top level of the gz menu. This allows quicker retries whenever you want instead of having to run to the exit first.

image

### DIFF
--- a/src/p2gz/segmentHistory.cpp
+++ b/src/p2gz/segmentHistory.cpp
@@ -54,7 +54,7 @@ void SegmentHistory::update()
 	WarpDestination current_dest = current_segment->dest;
 	const u32 btn                = p2gz->controller->getButtonDown();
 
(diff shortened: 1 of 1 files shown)


src/p2gz/segmentHistory.cpp
```diff
@@ -46,13 +46,15 @@ const Segment* SegmentHistory::cur_segment()
...
+	const Segment* current_segment = cur_segment();
+	if (!current_segment) {
+		return;
+	}
 
-		const u32 btn = p2gz->controller->getButtonDown();
+	WarpDestination current_dest = current_segment->dest;
+	const u32 btn                = p2gz->controller->getButtonDown();
 
+	if (entering_next_sublevel || p2gz->menu->is_root_open()) {
```
HeartPiece44: I think this needs a `&& in_cave_play()` check

PR #195, opened 2025-11-15
Add treasure counts to presets (for VoR first collection day end)

HeartPiece44:
Adds collected treasure counts to presets and feeds them to global play data as required. This fixes a bug where warping to an enter SCx preset would then end the day on first treasure collection, even though Courage Reactor should have been "collected" on Day 6.


include/p2gz/Preset.h
```diff
@@ -94,6 +97,7 @@ struct Preset {
 	bool plug_destroyed; // no more than one plug per level
 	bool apply_pokos;
 	u8 day;
+	u8 treasure_counts[4]; // indexed by course index
```
mayabyte: Shouldn't it be possible to derive this from the treasure spawn overrides rather than storing it separately? 4 bytes for each preset doesn't sound like a lot but if we can avoid them it'll help in the long run.

PR #197, opened 2025-11-15
Enable quick-retry for AG segments

mayabyte:
AG segments can now be retried just like cave segments. There's only one basic retry option as same-seed and from-beginning options don't make sense here.

Additional changes:
- Properly free old `Segment`s when the ringbuffer reaches capacity
- Use a proper type for cutscene bitflags in presets instead of two u32s
- Populate upgrades and cutscene flags in generated presets using actual current game state rather than the previous preset
- Begin `Segment`s in the Load state to more closely match RTA segments and to ensure one is made for AGs as well

### DIFF
--- a/src/p2gz/ekEditor.cpp
+++ b/src/p2gz/ekEditor.cpp
@@ -71,13 +71,13 @@ void EKEditor::check_upgrades()
 	set_menu_opt(OlimarData::ODII_JusticeAlloy, "metal suit Z");
 }
 
-u16 EKEditor::get_upgrades_bitfield()
+BitFlag<u16> EKEditor::get_upgrades_bitfield()
 {
-	u16 bitfield = 0;
+	BitFlag<u16> bitfield;
 	for (size_t i = Game::OlimarData::ODII_FIRST_EXPLORATION_KIT_ITEM; i < Game::OlimarData::ODII_LAST_EXPLORATION_KIT_ITEM; i++) {
 		const u16 mask = 1 << i;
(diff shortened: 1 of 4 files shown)


include/p2gz/CutsceneToggle.h
```diff
@@ -99,6 +99,42 @@ struct CutsceneToggle {
...
+			bf1 |= (1 << flag);
+		} else {
+			bf2 |= (1 << (flag - 32));
+		}
+	}
+
+private:
+	u32 bf1;
+	u32 bf2;
+};
```
jonathonbrady: Do we need a custom type? Why not use `BitFlag<u64>`?

src/p2gz/ekEditor.cpp
```diff
@@ -71,6 +71,18 @@ void EKEditor::check_upgrades()
...
+{
+	u16 bitfield = 0;
+	for (size_t i = Game::OlimarData::ODII_FIRST_EXPLORATION_KIT_ITEM; i < Game::OlimarData::ODII_LAST_EXPLORATION_KIT_ITEM; i++) {
+		const u16 mask = 1 << i;
+		if (playData->mOlimarData->hasItem(i)) {
+			bitfield |= mask;
+		}
+	}
+	return bitfield;
+}
```
jonathonbrady: Can you return a `BitFlag<u16>` instead?

HeartPiece44 (approved):
Nice work! Code looks good, functions as expected 🚢

PR #218, opened 2025-11-17
add fast empress trainer

jonathonbrady:
Adds a new trainers list menu that currently only contains a fast empress trainer. Automatically restarts on success and failure, displays the current flick count and its valid range, along with the first damage frame in the sleeping animation cycle.

I want to refactor some of this code into a trainer manager, but I don't think we'll know what that should look like until we implement a few more, so I've just hardcoded some things for now and will clean them up later.

### DIFF
--- a/src/sysGCU/system.cpp
+++ b/src/sysGCU/system.cpp
@@ -926,13 +926,6 @@ bool System::beginFrame()
 {
 	mCpuRetraceCount = 0;
 	JUTGamePad::read();
-
-	// @P2GZ - main update and menu drawing
-	if (p2gz) {
-		p2gz->update();
-		p2gz->draw_2d();
-	}
-
 	mDvdStatus->update();
 }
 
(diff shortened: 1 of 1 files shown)


HeartPiece44 (approved):
INSANELY rad, feels really slick and the information on success/failure is crazy helpful. Agreed on refactoring once we have enough trainers to make it worth it, definitely not a priority for this second.

Only comment is that the frame text line is a bit hard to read, since it's printed over the top of the Pikmin counts. I'm not gonna block on that because I have no good suggestions on where else to put it, and it's legible enough at the end on fadeout at the very least, but if you find a way to fix that that'd be mint.

HeartPiece44:
Looks 10 times better, I don't think it's in the way at all, especially since the text comes up after you've already c-sticked Pikmin on her, so it's not blocking any visual cues. Thanks for doing it!

Issue #205, opened 2025-11-16
Warping to an AG with collision viewer active causes lots of Z-fighting


mayabyte:
**Description of the bug**
Warping to an above-ground area with the collision viewer active causes lots of Z-fighting

<img width="1477" height="1088" alt="Image" src="https://github.com/user-attachments/assets/4fc8679c-820d-4f58-8618-992e3fb6948d" />

**Steps to Reproduce**
1. Warp to a cave
2. Turn on collision viewer
3. Warp to an AG
4. Observe Z-fighting

It can be fixed by disabling and re-enabling collision viewer.

PR #219, opened 2025-11-17
fix: reset collision viewer on warp

jonathonbrady:
Closes https://github.com/p2gz/p2gz/issues/205.

### DIFF
--- a/include/p2gz/StructureEditor.h
+++ b/include/p2gz/StructureEditor.h
@@ -13,12 +13,20 @@ struct StructureEditor {
 	struct GateWrapper {
 		GateWrapper()
 		{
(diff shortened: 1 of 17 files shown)


HeartPiece44 (changes requested):
Z-fighting seems gone! However, the collision viewer seems to be on by default and un-synced from the menu option. Can be fixed by toggling the option twice to re-sync it, but yeah.

(Boot up game -> warp to AG or cave from file select -> collision viewer is on -> check menu option for collision viewer -> says false)

jonathonbrady:
Good catch, I didn't even test warping with collision viewer disabled. Should be fixed now.

PR #221, opened 2025-11-18
Implement EK and cave discovery cutscene presets and options

### LINKED ISSUE #196: Add cave discovered cutscenes to cutscene editor and presets
Currently cave discovery cutscenes play at inappropriate times.

HeartPiece44:
Closes #196 by adding exploration kit discovery cutscenes and cave hole discovery cutscenes to the cutscene menu, as well as to presets as appropriate. 

Also renames the HelperInlines.h header to Utility.h and adds/adjusts some helper functions/enums around caves and areas.

### DIFF
--- a/include/p2gz/Utility.h
+++ b/include/p2gz/Utility.h
@@ -60,6 +60,7 @@ inline CourseIndex get_area_from_cave(CaveIndex cave)
 
 	default:
 		OSReport("[P2GZ WARN] get area from cave: bad cave id [%d]\n", cave);
+		GZASSERTLINE(false); // force crash
 		return COURSE_WW;
 	}
 }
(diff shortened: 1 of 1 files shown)


include/p2gz/Utility.h
```diff
@@ -9,6 +9,102 @@
...
+		return COURSE_PP;
+
+	case CAVE_CoC:
+	case CAVE_HoH:
+	case CAVE_DD:
+		return COURSE_WW;
+
+	default:
+		OSReport("[P2GZ WARN] get area from cave: bad cave id [%d]\n", cave);
+		return COURSE_WW;
```
jonathonbrady: Thoughts on `GZASSERTLINE(false)` instead? I prefer blowing up when we reach theoretically unreachable code rather than returning an arbitrary value.

HeartPiece44:
Good idea - I think both is good for now, so we have info to debug with, but you're right - it should force a crash if we call those functions with bad values. Added!

Issue #220, opened 2025-11-18
Captain Health not displaying upon warping


SgtP2:
https://gyazo.com/4e47d5d344029b772ba18214133b9796

upon loading into scx1 from file select with the preset, adjusting the squad, then warping again with no preset, captain health bars do not display and they die in 1 hit

PR #222, opened 2025-11-18
Several bug fixes and improvements

mayabyte:
- `Game::PikiContainer` now has a destructor so using it as a local variable in a function body doesn't leak memory
- SquadEditor now uses PikiContainer instead of a vec of types
- Fixes a crash caused by an incorrect bound in an assert in `gz::RingBuffer`
- Captains now have full health when warping in all circumstances, even when no preset is applied (closes https://github.com/p2gz/p2gz/issues/220)
- Preset menu code has been moved to a separate cpp file for cleanliness
- The warp menu now shows you the name of your currently selected preset
- Changing your squad with the squad editor now sets your current warp preset to nullptr so you don't accidentally override it when warping


HeartPiece44 (approved):
Looks good! Helluva lot of fixes for annoying stuff, glad to see it. Seems to function well in the cases I can think of testing.

The GZEXPECT macro is a great addition, will be keen to use that in more places in the future.

PR #227, opened 2025-11-19
Fix: drastically cut DOL size by un-inlining iterator functions

HeartPiece44:
Cuts the size of the DOL by a fair chunk (without changing or removing any functionality) by forcing `Iterator` functions to not inline. This creates about an extra ~110KB of space in the game heap, which we could siphon to the system heap for our purposes if we want to down the line. 

Credit to @PikHacker for the idea - thank you!

### DIFF
--- a/include/Iterator.h
+++ b/include/Iterator.h
@@ -97,7 +97,9 @@ struct Iterator {
 	 *
 	 * @return Reference to the incremented iterator.
 	 */
-	inline Iterator<T>& operator++()
+	// @P2GZ: un-inline iterator functions
+	// inline Iterator<T>& operator++()
+	Iterator<T>& operator++()
 	{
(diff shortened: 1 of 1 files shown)


include/Iterator.h
```diff
@@ -101,9 +109,13 @@ struct Iterator {
 	 *
 	 * @return true if the current element satisfies the condition, false otherwise.
 	 */
-	inline bool satisfy() { return mCondition->satisfy(mContainer->get(mIndex)); }
+	inline bool satisfy()
```
jonathonbrady: Should we remove `inline` here? (At least for semantics? I don't actually know how forceful `FORCE_DONT_INLINE` is.)

HeartPiece44: We should, nice catch - have removed them now.

PR #233, opened 2025-11-20
Squad Editor improvements and polish

mayabyte:
- Squad editor now shows and can edit all pikmin on the field, not just the ones in your squad
  - It shows wild pikmin but they're not editable. TBD what the desired behavior is.
- Colors have been reordered to the one most people probably expect (red, yellow, blue, purple, white, bulbmin)
- Stages have been reordered to put the most common ones first (flower, leaf, bud)
- Added a new "clear all" option in the "pikmin" menu that kills all non-wild pikmin on the field. Useful for completely replacing your squad without having to lower all the options to 0 manually.
- `RangeMenuOption` with the `Wrap` mode has had its behavior improved when using the 10x functionality: the range now saturates at the min and max values so you can easily reach the highest and lowest values. Pressing dL or dR respectively when on the extremities still wraps as before.
- Squad editor code has been refactored to use a sync callback instead of an update loop
- Menu definition for the squad editor options has been moved into `SquadEditor::init` to declutter `GZMenu::init`
- Each editor option's callback is now a `CurriedDelegate2` with the right kind and stage arguments to avoid most of the menu traversal
- Fixed a bug where calling `SquadEditor::kill_piki` with a count <= 0 would kill all pikmin of that type

image


HeartPiece44 (approved):
Nice, good catch. Looks good now, seems to work as expected. 🚢

PR #235, opened 2025-11-22
Implement AT above-ground presets (+ scaffold draft of cave presets)

HeartPiece44:
Sets up presets for All Treasures for above ground splits. Also refactors PoD and AT presets into separate cpp files to declutter presetMgr.cpp.

Also sets up cave presets for all All Treasures caves, but leaves these commented out. With them included, the code size increases approximately 40KB and requires the system heap to be expanded by approximately 40KB as well to hold them in memory when initialised. The combination of the two causes the game heap to be too small in WW day 31 and crashes the game, but does manage to load in AW and similar, so it's pretty close. I think to implement them without tying our hands too tightly, we may need to put them in text files and parse them either on p2gz init or on preset application, to reduce the code burden (and maybe sys heap burden). Probably something to dig into more for the next alpha release rather than the current one.

Also worth noting - the AT presets are implemented as multiple functions, one for each area - this is because if they're all stacked inside the one function, it causes the compile time to go up by 6-8 minutes (!!), while only saving a relatively small amount of code size (~10KB). Keeping all the PoD presets inside the one function is fine however, and only adds 5s to compile time while saving 2KB of code size.

### DIFF
--- a/src/p2gz/presetsAT.cpp
+++ b/src/p2gz/presetsAT.cpp
@@ -18,19 +18,33 @@ void PresetMgr::init_at_vor1_presets()
 {
 	const char* g_at_post_enter_ec[] = { "EC gate" };
 	const char* b_post_enter_ec[]    = { "landing area bag (15)", "hubcap bag (35)" };
-	// presets.push((new Preset("Day 1", AT))->set_day(1));
+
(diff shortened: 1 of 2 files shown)


mayabyte (approved):
Fantastic work. Getting the presets moved to text files will be a good thing to do for next version, but it'll be good to have the AGs in to start with.

PR #246, opened 2025-11-23
Fix: Pikmin no longer recalculate their paths mid-carry

### LINKED ISSUE #242: Pikmin recalculate their paths during carry in certain places
**Description of the bug**
When napsacking in certain places, the pikmin carrying the captain will recalculate their path (and goal) despite no pikmin being added or taken away from the carry.

**Steps to Reproduce**
1. Warp to AW using enter BK preset.
2. Perform BK stump napsack.
3. Observe pikmin re-calculating the route while going up the root, making them pause and often changing the destination onion.

**Crash Log or Screenshots**
https://www.twitch.tv/jpeppersons/clip/EasyLaconicMonkeyPastaThat-kbywBMoj_XDcScK5

HeartPiece44:
Fixes an error when copying from the decomp repo in aiPrimitives.cpp that caused Pikmin to recalculate their paths during a carry when they shouldn't be - file should be properly equivalent now. Closes #242.

### DIFF
--- a/src/plugProjectKandoU/aiPrimitives.cpp
+++ b/src/plugProjectKandoU/aiPrimitives.cpp
@@ -1174,6 +1174,8 @@ bool ActPathMove::contextCheck(int idx)
 		if (len > 700.0f) {
 			return false;
 		}
+
+		return true;
 	}
 
 	Vector3f point           = crGetPoint(idx);


mayabyte (approved):
👉 🚢

PR #249, opened 2025-11-24
Read Presets from files

mayabyte:
Reads presets from files to save on memory and binary size. 

- Each preset is to be defined in a separate file. 
  - At build time, the `files/presets` folder is read and an `all_presets.txt` file is generated with a list of all preset files. This serves as the master list of presets to load. 
  - At init time, all the preset files are read and a `PresetPreview` is constructed with just the name, category, and squad info so it can be shown in the menu. 
  - When the preset actually needs to be used for warping, the full thing is read from file at that point.
- Presets have been given a ref count so they can be freed at the appropriate time even when being shared between Segments and the warp menu, etc.
  - A future improvement to save more memory could be to always deallocate presets that originate from file since they can always be retrieved again easily. This would require swapping between presets and preset previews dynamically in segments, so it's not a trivial change.
- Structure overrides now have a dedicated struct and are defined by area and position, with a new data field for defining things like number of stages remaining on gates and bridges.
- Presets now have an origin defined based on where they came from: file, memcard (not yet implemented but I'm aspirational), or generated. 

ALL presets have been ported to files, including the commented out AT presets.

### DIFF
--- a/include/p2gz/Preset.h
+++ b/include/p2gz/Preset.h
@@ -119,8 +119,7 @@ struct Preset {
 
 public:
 	Preset();
-	Preset(const char* name_, PresetCategory category_);
-	Preset(Preset& other);
+	Preset(Preset& other) { GZEXPECT(false, "do not use Preset copy ctor"); }
 	~Preset();
 
 	void read_file(const char* filename);
--- a/src/p2gz/preset.cpp
+++ b/src/p2gz/preset.cpp
@@ -76,85 +76,6 @@ Preset::Preset()
 	onion_pikis.clear();
 }
 
-Preset::Preset(const char* name_, PresetCategory category_)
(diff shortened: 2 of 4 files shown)


mayabyte:
Now ready for review

HeartPiece44 (changes requested):
Awesome! Fantastic work - insane to get all the AT presets working with so much heap leftover, even with the adjustment. Preset previews seem good now - however, I'm getting a crash warping back to an area after killing an enemy when it needs to be respawned (such as AW with enter BK preset -> kill bulborbs -> AW with enter BK preset) - I think the `get_preset_during_warp()` function assumes it's only being called while warping, while it's actually being called regardless of warp situation and then handling the result depending on warp.

Crash logs:

image

image

PR #250, opened 2025-11-26
fix: poll empress flick count for an additional frame

jonathonbrady:
Currently, we only poll empress flick count in state `QUEEN_Sleep`. This means the final flick count is always off by a frame, which we didn't notice before because it's only obviously wrong when the flick count crosses 256 or 305 during the frame her state changes. This updates the empress trainer to poll for an additional frame when her state changes to `QUEEN_Damage` or `QUEEN_Flick`.


mayabyte (approved):
👉 🚢

PR #261, opened 2025-11-30
Onion editor

mayabyte:
Adds an editor for onion pikmin counts. Mostly made by @jonathonbrady.

The code for an onion unlock/move feature is included in this PR, but using it repeatedly triggers a persistent crash related to allocating the onion beam particle effect. None of us have been able to solve it so we decided to ship this with it commented out for now.


HeartPiece44 (approved):
Nice. Seems to work well, thanks for fixing it up - we can revisit the moving onions thing later if we get demand for it - I might ask some of the more particle-minded modders about how to handle particle effects better if/when we do.

PR #263, opened 2025-12-01
Fix crash when saving after warping from file select

mayabyte:
Now emulates having no memcard when attempting to save after warping from file select.

### DIFF
--- a/src/p2gz/warp.cpp
+++ b/src/p2gz/warp.cpp
@@ -304,15 +304,8 @@ void Warp::do_warp()
 		Game::naviMgr->clearDeadCount();
 	}
 
-	// Disable the option for skipping save prompts and force it to be on
-	// when warping out of the menu. This is because there's no save file selected,
-	// so saving in this state causes a crash.
 	if (warping_from_menu) {
-		p2gz->in_save_file                      = false;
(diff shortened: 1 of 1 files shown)


src/p2gz/warp.cpp
```diff
@@ -308,6 +308,7 @@ void Warp::do_warp()
 	// when warping out of the menu. This is because there's no save file selected,
 	// so saving in this state causes a crash.
 	if (warping_from_menu) {
+		p2gz->in_save_file                      = false;
 		ToggleMenuOption* skip_save_prompts_opt = static_cast<ToggleMenuOption*>(p2gz->menu->get_option("settings/skip save prompts"));
 		skip_save_prompts_opt->visible          = false;
 		skip_save_prompts_opt->set_selection(true);
```
HeartPiece44: Do we still need to force the save skipping to be active when warping from file select with this solution? Not sure who's going to be turning it off, but maybe doesn't make sense to make it inaccessible anymore.

HeartPiece44 (approved):
Looks good! One comment re: save skipping option, but aside from that seems to work beautifully, nice fix.

src/p2gz/warp.cpp
mayabyte: Good point, I'll remove it

PR #267, opened 2025-12-03
Fix release scripts and a console bug after preset files PR

mayabyte:
- Fixes an issue only present when playing from a built ISO where the all-presets file was being read past the end
- Actually includes the all-presets file in the release folder when making the release zip
- Removes all-presets.txt from git since it's auto-generated at build time and shouldn't be manually edited.


HeartPiece44 (approved):
🚢

Issue #251, opened 2025-11-26
VoR aboveground pellets/eggs do not respawn properly


JpepWasTaken:
**Description of the bug**
On days 3 and 4, pellet posies and eggs do not respawn properly if the user warps back into the aboveground after collecting them or breaking them. 

**Steps to Reproduce**
1. Go to VoR AG using the day 4 preset
2. Collect the 5 pellet near spawn or break the egg push egg
3. Warp to the same area again. The pellet posy and egg are missing.

This concerns both 1 pellets on day 3, all 5 and 10 pellets on day 4 as well as the eggs.

Issue #256, opened 2025-11-26
[Feature] Full PoD presets


JpepWasTaken:
It would be nice to have a completely thorough, full PoD preset list. At the moment we are missing Day 2 (due to technical concerns, I believe), Enter WFG and Enter FC. Enter WFG was recently rerouted - even though it was not conclusive I believe this speaks to usefulness of thorough preset-making.

PR #274, opened 2025-12-19
Load nonloop gens in a disabled state so presets can manipulate them

mayabyte:
- Always loads all nonloop generators prior to the current day so they can be force-spawned by presets. They're marked as disabled so they don't spawn anything on their own and they aren't saved in subsequent non-warp loads.
- Corrects a mistake in the Day 2 preset; the bulborb is now correctly marked to spawn
- Makes the first pluck cutscene play as it should on day 2

Closes https://github.com/p2gz/p2gz/issues/251 and https://github.com/p2gz/p2gz/issues/256.


HeartPiece44 (approved):
🚢

PR #279, opened 2026-06-06
Add "stuck pikmin" display, "treasure velocity" display

JpepWasTaken:
Adds options to display the number of pikmin stuck to an enemy under the "enemy debug info" menu. Adds option to display the velocity of a treasure under the "treasure debug info" menu. Refactors the "treasure debug info" menu - it used to be a straightforward true/false toggle, now there is a toggle for displaying each piece of information.


src/p2gz/enemyDebugInfo.cpp
```diff
@@ -278,6 +278,10 @@ void EnemyDebugInfo::draw_enemy_dbg(Game::EnemyBase* enemy, Graphics* gfx)
 		draw_timer(enemy, gfx, info, pos);
 		info.mPerspectiveOffsetY += line_height;
 	}
+	if (draw_stuck_pikmin) {
+		gfx->perspPrintf(info, pos, "stuck: %d", (int)enemy->mStuckPikminCount);
```
HeartPiece44, line 282: Note: using just `mStuckPikminCount` will include pikmin stuck to mouth parts (i.e. pikmin being swallowed) - that's probably fine, since I assume this is mostly worth knowing for other stuff that also uses `mStuckPikminCount`, but worth keeping in mind.

src/p2gz/treasureDebugInfo.cpp
```diff
@@ -57,9 +61,23 @@ void TreasureDebugInfo::draw_treasure_dbg(Game::Pellet* pellet, Graphics* gfx)
...
+	if (draw_carry_power) {
+		gfx->perspPrintf(info, pos, "carry power: %.1f", pellet->mCarryPower);
+		info.mPerspectiveOffsetY += 22;
+	}
+	if (draw_velocity) {
+		Vector3f speed;
+		speed.x = pellet->getVelocity().x;
+		speed.y = pellet->getVelocity().y;
+		speed.z = pellet->getVelocity().z;
+		gfx->perspPrintf(info, pos, "velocity: %.1f", speed.sqrMagnitude() / 100.0f);
```
HeartPiece44: Should this be `speed.length()` rather than `speed.sqrMagnitude() / 100.0f`? Unless there's another reason to do this adjustment (map units vs code units etc, but I still think that'd be an adjustment on the root rather than the squared magnitude)? 

Also, it might be worth having a check on the pellet state or a call to `isPicked()` or something to zero this out when it's not Actually Moving - stationary treasures atm have like 0.4 speed or something, it's just the state is such they don't move. Open to leaving it as-is if you think it's useful info.

src/p2gz/gzmenu.cpp
```diff
@@ -126,7 +127,13 @@ void GZMenu::init_menu()
 				->push(new ToggleMenuOption("draw current state", true, new Delegate1<NaviDebugInfo, bool>(p2gz->navi_debug_info, &NaviDebugInfo::set_draw_state_enabled)))
 				->push(new ToggleMenuOption("show captain swap time", true, new Delegate1<NaviDebugInfo, bool>(p2gz->navi_debug_info, &NaviDebugInfo::set_draw_swap_time_enabled)))
 			))
-			->push(new ToggleMenuOption("treasure debug info", false, new Delegate1<TreasureDebugInfo, bool>(p2gz->treasure_debug_info, &TreasureDebugInfo::set_enabled)))
+			->push(new OpenSubMenuOption("treasure debug info", (new ListMenu())
+				->push(new ToggleMenuOption("enable", false, new Delegate1<TreasureDebugInfo, bool>(p2gz->treasure_debug_info, &TreasureDebugInfo::set_enabled)))
+				->push(new ToggleMenuOption("draw name & id", true, new Delegate1<TreasureDebugInfo, bool>(p2gz->treasure_debug_info, &TreasureDebugInfo::set_draw_name_id_enabled)))
```
HeartPiece44: Nit: can you change the & to "and" so Jon doesn't have a conniption

HeartPiece44 (commented):
Hell yeah jpep, nice work - appreciate you making the treasure debug info aligned with the enemy debug info setup, and it's cool having the extra counts, especially the treasure speed. Have added a couple comments, but happy to approve it once they're addressed and/or adjusted 🚢

src/p2gz/enemyDebugInfo.cpp
JpepWasTaken: Yep, I thought this was potentially interesting info too since certain glitches interact with this. For instance if a Snitchbug picks up a purple pikmin, it counts as stuck onto it, and so hitting it with any other color will force a down --- I think this behaviour makes a little bit more sense if the stuck count properly displays a >0 number when the Snitchbug is flying with a purple in hand.

src/p2gz/treasureDebugInfo.cpp
JpepWasTaken: Oops! You're right, I misread sqrMagnitude as the square root of magnitude, and got misdirected on the /100f thing since i happened to test it on a treasure with a speed that maxes out at exactly 100. Good catch!

My preference would be to keep the 0.4 something speed for 2 reasons:
1. sometimes treasures slide or are pushed and I'd like this speed to be caught by the function too 
2. i believe the way captain speed display was implemented was to display that constant speed too, so it would align with precedent

HeartPiece44 (approved):
Thanks for the fixes - looks good!

PR #280, opened 2026-06-09
fix: add all_presets.txt

jonathonbrady:
For no discernible reason, this file was added to `.gitignore` and never committed to `main`, despite being required to boot.


HeartPiece44 (approved):
LGTM

PR #281, opened 2026-06-09
fix: first corpse in cave CS only plays once

JpepWasTaken:
Fix issue where the first corpse cutscene only plays once by setting the isGot flag to true if the enemy is a snowy bulborb or a sheargrub and the cutscene has not played yet.

Edge case: mitites on HOB4 as first bug, or manually using the gz menu to collect an exotic enemy as first corpse twice in a row.

### DIFF
--- a/src/plugProjectKandoU/pelletState.cpp
+++ b/src/plugProjectKandoU/pelletState.cpp
@@ -225,6 +225,13 @@ bool PelletGoalState::checkMovie(Pellet* pelt)
 	// onion above ground
 	if (gameSystem->isStoryMode()) {
 		isGot = playData->firstCarryPellet(pelt);
+		// @P2GZ: fix issue where the first corpse cutscene only plays once 
+		EnemyBase* enemy = static_cast<EnemyBase*>(pelt->mPelletView->mCreature);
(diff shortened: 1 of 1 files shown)


src/plugProjectKandoU/pelletState.cpp
```diff
@@ -225,6 +225,13 @@ bool PelletGoalState::checkMovie(Pellet* pelt)
 	// onion above ground
 	if (gameSystem->isStoryMode()) {
 		isGot = playData->firstCarryPellet(pelt);
+		// @P2GZ: fix issue where the first corpse cutscene only plays once 
+		EnemyBase* enemy = static_cast<EnemyBase*>(pelt->mPelletView->mCreature);
+		if (enemy != nullptr && (enemy->getEnemyTypeID() == EnemyTypeID::EnemyID_YellowKochappy || enemy->getEnemyTypeID() == EnemyTypeID::EnemyID_UjiA || enemy->getEnemyTypeID() == EnemyTypeID::EnemyID_UjiB) && !playData->isDemoFlag(DEMO_First_Corpse_In_Cave)) {
```
HeartPiece44: I don't think we need the UjiA (female sheargrub) check here - we'll always collect one in the enter HoB or enter WFG above ground, and there aren't any in EC, so it should never set off the cutscene even in AT.

HeartPiece44 (commented):
Nice work! Quick elegant fix, far simpler than I was expecting, that's great. I think the female sheargrub check should be removed, but aside from that, looks good and works perfectly.

src/plugProjectKandoU/pelletState.cpp
JpepWasTaken: Good catch!

HeartPiece44 (approved):
LGTM, nice catch on the female sheargrub check!

PR #284, opened 2026-06-15
collision viewer enhancements

jonathonbrady:
Draws edges between collision triangles and reduces the opacity of triangles that occlude the captain.
image
image

### DIFF
--- a/include/p2gz/CollisionViewer.h
+++ b/include/p2gz/CollisionViewer.h
@@ -3,6 +3,7 @@
 
 #include <Game/mapMgr.h>
 #include <Game/mapParts.h>
+#include <Color4.h>
 #include <Sys/Edge.h>
 #include <Sys/TriIndexList.h>
 #include <Sys/Triangle.h>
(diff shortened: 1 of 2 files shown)


HeartPiece44 (approved):
This is so unbelievably cool - it looks great and works so well now. Far more professional than it was, super slick. Went through it thoroughly + play-tested, LGTM.

HeartPiece44 (changes requested):
Actually, messing around with it more, I found a crash (sorry):
Steps to reproduce:
1. Warp to AW with e.g. enter HoB preset
2. Turn on collision viewer and waypoint viewer
3. Walk towards the black gate (on the way to the bag/WFG/HoB)
4. Game freezes, then crashes after a few seconds.

Crash log:

Image

Image

Image

jonathonbrady:
The waypoint viewer might have been a red herring. I made the above fix and couldn't reproduce the crash.

HeartPiece44 (approved):
Sick - works great now, thanks!

PR #285, opened 2026-06-18
add early blues trainer

jonathonbrady:
Adds Early Blues to the trainers menu.

1. Enables an adjustable inset camera that shows the out of bounds collision.
2. Resets the captain after each attempt back to the initial starting position. Players can adjust this position by pressing Y if they just want to practice, e.g., the napsack timing, but don't want to repeat the initial movement. The position can be reset back to inside the scale by holding B.
3. Provides feedback on each attempt, including whether the captain napped in time, reached the correct waypoint, or would have softlocked in vanilla.

https://github.com/user-attachments/assets/fd0dda4a-9dfc-47fa-a8f4-f764a287bbe1


HeartPiece44 (changes requested):
This is honestly so sick, and I'm so excited to have it in the hack, I think people are going to go fucking nuts for it. Really great work. Unfortunately, found a couple bugs:

Bug 1: if you try and enter the EB trainer from file select, nothing happens - hitting A does nothing. You can enter the fast empress trainer from file select correctly, however.

Bug 2: if you warp to an area with a preset "earlier" than the intended early blues split (such as pod enter HoB) and then enter the EB trainer, you will not have the napsack. entering EB trainer should probably give you the correct preset, or at least parts of it (including resetting the discover blue onion cutscene).

HeartPiece44 (approved):
Works perfectly now - super excited for this. LGTM

PR #287, opened 2026-06-18
Implement custom setting saving to memory card

HeartPiece44:
Settings from P2GZ can now be saved to the memory card. This is implemented with a second data file on the memory card, separate from the vanilla one, and is versioned such that adding new features will not break old save files. 

Currently, the capacity of the new data is set to 16 blocks, which is 128kb of serialised data. If we think we would ever need more than this, we should bump the size now however.

As a starting point, the initial version saves the the skippable cutscenes, skipping save prompts, timer enabled and subtimer enabled bools, as well as the treasure region setting, so they'll persist across soft and hard resets. We can extend these whenever we want, and instructions for how to do so are in the Settings.cpp/Settings.h files.


include/p2gz/Settings.h
```diff
@@ -0,0 +1,80 @@
...
+// the version it first appeared in, so saves stay compatible across p2gz releases:
+//   - older save + newer p2gz: blocks the save predates keep their setDefault() values
+//   - newer save + older p2gz: trailing blocks the build doesn't know are simply not read
+// Only ever append; bump + branch (never reorder/resize an existing field) on layout change.
+enum SettingsVersion {
+	SETTINGS_VERSION_V1 = 1, // bool-settings array (count-prefixed) + treasure region
+
+	// ---- add new SETTINGS_VERSION_<name> entries above this line ----
+	SETTINGS_VERSION_CURRENT = SETTINGS_VERSION_V1, // point this at the newest entry
+};
```
jonathonbrady: Should we call this something other than `Settings`, since we are likely to store arbitrary data on the memory card beyond the player's preferences?

include/Game/MemoryCard/Mgr.h
```diff
@@ -42,6 +42,16 @@ enum MemoryCardMgrFlags {
...
 struct Mgr : public MemoryCardMgr {
+	// @P2GZ: settings on mem card
+	// layout of dedicated settings file (see cP2GZFileName)
+	// the file is pre-sized with spare data capacity so the written payload can grow later
+	// - to increase size, change P2GZ_PAYLOAD_SIZE from 0x2000 up to a max of 0x1E000
+	enum P2GZDataLayout {
+		P2GZ_FILE_SIZE    = 0x20000, // total allocation: 0x2000 header + 0x1E000 data capacity (16 blocks)
+		P2GZ_DATA_OFFSET  = 0x2000,  // payload begins right after the header block
+		P2GZ_PAYLOAD_SIZE = 0x2000,  // bytes currently written/read/checksummed (may grow to <= 0x1E000)
+	};
```
jonathonbrady, line 53: Can we expand `P2GZ_PAYLOAD_SIZE` at any time? If not, should we preemptively make it as large as possible now?

src/p2gz/settings.cpp
```diff
@@ -0,0 +1,111 @@
...
+#include "System.h"
+#include "stream.h"
+
+using namespace gz;
+
+// Identifies our memory card payload. If the leading word doesn't match, the buffer isn't
+// ours (uninitialised/foreign/corrupt) and we fall back to defaults rather than read garbage
+#define SETTINGS_MAGIC 'p2gz'
+
+void Settings::setDefault()
```
jonathonbrady: nit: snake case

src/sysGCU/pikmin2MemoryCardMgr.cpp
```diff
@@ -253,6 +258,106 @@ bool Mgr::loadGameOption()
...
+}
+
+// @P2GZ: settings on mem card
+// queue up reading our settings file (used once at boot)
+bool Mgr::loadP2GZData()
+{
+	// use a new custom command id (18 for load, 17 for save)
+	setCommand(18);
+	return true;
+}
```
jonathonbrady, line 289: Can we make these magic numbers (5, 17, and 18) `const`s?

src/sysGCU/pikmin2MemoryCardMgr.cpp
```diff
@@ -451,6 +556,18 @@ bool Mgr::doCardProc(void*, MemoryCardMgrCommand* command)
...
+	// save and load commands for our settings file
+	case 17:
+		setFlag(MCMFLAG_IsWriting);
+		result = commandSaveP2GZData();
+		resetFlag(MCMFLAG_IsWriting);
+		break;
+
+	case 18:
+		result = commandLoadP2GZData();
+		break;
```
jonathonbrady, line 569: Magic numbers should be replaced with `const`s here, too.

src/p2gz/gzmenu.cpp
```diff
@@ -144,15 +144,15 @@ void GZMenu::init_menu()
...
 			->push(new PerformActionMenuOption("reset", new Delegate<Timer>(p2gz->timer, &Timer::on_reset)))
 		))
 
 		// General game behaviors and options for how the gz menu looks and behaves
 		->push(new OpenSubMenuOption("settings", (new ListMenu())
-            ->push(new ToggleMenuOption("skippable cutscenes", true, new Delegate1<SkippableCutscenes, bool>(p2gz->skippable_cutscenes, &SkippableCutscenes::toggle_skippable)))
-            ->push(new ToggleMenuOption("skip save prompts", true, new Delegate1<SkipSave, bool>(p2gz->skip_save, &SkipSave::toggle_save_skip)))
+            ->push(new ToggleMenuOption("skippable cutscenes", p2gz->settings->bool_settings[SETTING_cutscenes_skippable], new CurriedDelegate1<Settings, SettingId, bool>(p2gz->settings, &Settings::set_bool, SETTING_cutscenes_skippable)))
+            ->push(new ToggleMenuOption("skip save prompts", p2gz->settings->bool_settings[SETTING_skip_save_prompts], new CurriedDelegate1<Settings, SettingId, bool>(p2gz->settings, &Settings::set_bool, SETTING_skip_save_prompts)))
```
jonathonbrady: I'm worried we might find saving in real time to be much less performant on console, given that memory card IO is way slower than on emulator. It's probably bad UX if the game just freezes for a few seconds (or whatever it will do) every time you update these settings, though maybe that's acceptable since these are unlikely to be updated often.

jonathonbrady (changes requested):

include/Game/MemoryCard/Mgr.h
HeartPiece44: We should be able to expand `P2GZ_PAYLOAD_SIZE` whenever we like, up to a max of `0x1E000` (bounded by `P2GZ_FILE_SIZE - P2GZ_DATA_OFFSET`). If we ever want to increase `P2GZ_FILE_SIZE` we have to do that now though (or invalidate every user's P2GZ save data later).

src/p2gz/gzmenu.cpp
HeartPiece44: Good point, hadn't considered that. I think we leave it for the second, and then pop it on the list of console checks for testers. If it ends up being hilariously bad, we adjust and lump it into a like "save to card" button within the menu or something.

HeartPiece44:
Have addressed comments and adjusted the naming. Will maybe need to adjust things once it gets tested on console, but we can relegate that to another PR later I think.

jonathonbrady (approved):

PR #291, opened 2026-06-20
Implement poko % cutscenes to relevant PoD abovegrounds

JpepWasTaken:
Percent cutscenes now play properly on every segment where they are expected to in Pay off Debt. 

Percent cutscenes are not yet implemented for AT, and it would probably be useful to make a menu option to turn them all off forcefully for segments in which they may or may not play (enter scx).

### DIFF
--- a/src/p2gz/preset.cpp
+++ b/src/p2gz/preset.cpp
@@ -301,7 +301,6 @@ void Preset::apply()
 	p2gz->warp->set_enter_area_type(enter_kind);
 
 	// set whether %cutscene should be forced to play or not
-	OSReport("right before setting enable percent cutscenes \n"); 
 	if (play_repay_demo){
 		p2gz->poko_editor->repay_demo_enabled = true; 
 	}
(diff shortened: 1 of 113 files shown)


src/p2gz/preset.cpp
```diff
@@ -299,6 +300,12 @@ void Preset::apply()
 
 	p2gz->warp->set_enter_area_type(enter_kind);
 
+	// set whether %cutscene should be forced to play or not
+	OSReport("right before setting enable percent cutscenes \n");
```
HeartPiece44: Nit: can we remove this OSReport?

src/p2gz/presetSerde.cpp
```diff
@@ -282,6 +287,8 @@ void Preset::read(Stream& input)
 
 	new_area_zoom.clear();
 	new_area_zoom.typeView = static_cast<u16>(input.readInt());
+	OSReport("Right before reading play_repay_demo \n");
```
HeartPiece44: Also this one?

HeartPiece44 (commented):
Nice! Thanks for getting this working - looks good for now, will make sure it works with the correct percentages once I add in pokos on another branch/PR. If you could remove a couple logs, that'd be great, then I think it's ready to merge.

HeartPiece44 (approved):
LGTM!

PR #292, opened 2026-06-21
Implement segment timer

JpepWasTaken:
This PR implements a segment timer that automatically measures the time elapsed between key actions, such as switching captains, killing enemies, starting a carry or breaking a gate. This will be a massive QoL increase for the nerds: no more need to open up vids and go through them frame by frame to tediously measure the length of a bunch of segments, it's all automatic now. 

The segment timer also comes with a toggle to keep in memory the best segments you've played in a given preset and to automatically display comparisons between your current play and your best. This will make it very easy to isolate great play and pick up on optimizations. 

There's potential to implement this with memory card saving so people can share their own best times in a segment for comparison (would go well with Jon's ghost idea)


include/p2gz/timer.h
```diff
@@ -93,13 +119,25 @@ struct Timer {
...
 	u32 skip_timer;  // for offsets because of P2GZ toggles
 	u32 pause_timer; // for when gz menu is open
 
 	u32 navi_swap_timer; // for measuring captain swap times
 
+	char curr_index; // tracks current active index of split_times
+	u32 split_times[20]; // tracks timestamps of trigger events 
+	u32 segment_times[20]; // tracks the segment times to display on screen 
+	u32 split_times_pause_lengths[20]; // position i tracks the pause length between split time i-1 and split time i 
+	u32 split_start_offset; // tracks e.g. comedowns
```
HeartPiece44: I don't think this is ever used anywhere outside of its setting function? Not sure of the purpose

include/p2gz/timer.h
```diff
@@ -73,6 +74,31 @@ struct Timer {
...
+	void set_split_on_gate_seg_enabled(bool enabled_) { split_on_gate_seg = enabled_; }
+	void set_split_on_gate_down_enabled(bool enabled_) { split_on_gate_down = enabled_; }
+	void set_split_on_bag_crush_enabled(bool enabled_) { split_on_bag_crush = enabled_; }
+	void set_split_on_poison_demo_enabled(bool enabled_) { split_on_poison_demo = enabled_; }
+	void set_split_on_carry_enabled(bool enabled_) { split_on_carry = enabled_; }
+	void set_split_on_enemy_death_enabled(bool enabled_) { split_on_enemy_death = enabled_; }
+	void set_draw_best_times_enabled(bool enabled_) { draw_best_times_enabled = enabled_; }
+	void set_draw_comparisons_enabled(bool enabled_) { draw_comparisons_enabled = enabled_; }
+	void reset_best_segments(); 
+	void enable_mark_run_to_discard() { OSReport("Calling enable_mark_run_for_discard \n");  mark_run_for_discard = true; }
```
HeartPiece44: Nit: can remove this extra OSReport

src/p2gz/gzmenu.cpp
```diff
@@ -148,6 +147,21 @@ void GZMenu::init_menu()
 			->push(new ToggleMenuOption("enabled", true, new Delegate1<Timer, bool>(p2gz->timer, &Timer::set_enabled)))
 			->push(new ToggleMenuOption("show sub-timer", false, new Delegate1<Timer, bool>(p2gz->timer, &Timer::set_sub_timer_enabled)))
 			->push(new PerformActionMenuOption("reset", new Delegate<Timer>(p2gz->timer, &Timer::on_reset)))
+			->push(new OpenSubMenuOption("segment timer", (new ListMenu())
+				->push(new ToggleMenuOption("enabled", true, new Delegate1<Timer, bool>(p2gz->timer, &Timer::set_segment_timer_enabled)))
+				->push(new ToggleMenuOption("draw best times", false, new Delegate1<Timer, bool>(p2gz->timer, &Timer::set_draw_best_times_enabled)))
+				->push(new ToggleMenuOption("draw comparisons", false, new Delegate1<Timer, bool>(p2gz->timer, &Timer::set_draw_comparisons_enabled)))
+				->push(new PerformActionMenuOption("reset best times", new Delegate<Timer>(p2gz->timer, &Timer::reset_best_segments)))
+				// this toggle below is not cancellable if it is accidentally pressed
+				->push(new PerformActionMenuOption("discard current run", new Delegate<Timer>(p2gz->timer, &Timer::enable_mark_run_to_discard)))
```
HeartPiece44, line 156: I think we should have a menu option that does a second confirmatory prompt for these - happy for that to be outside of scope for this PR and implement it later, but flagging it anyway.

src/plugProjectKandoU/pelletMgr.cpp
```diff
@@ -1871,6 +1867,12 @@ void Pellet::start_pmotions()
  */
 void Pellet::start_carrymotion()
 {
+	// @P2GZ - Segment timer 
+	// splits when a treasure starts to carry 
+	if (p2gz->timer->split_on_carry && (getKind() != PelletType::Treasure)){
```
HeartPiece44: Should this be `getKind() == PelletType::Treasure`? At the moment I think it splits on anything EXCEPT a treasure. Should also maybe make it `(getKind() == PelletType::Treasure || getKind() == PelletType::Upgrade)` so it handles the globe/EK treasures as well.

src/p2gz/timer.cpp
```diff
@@ -332,6 +431,56 @@ void Timer::cancel_navi_swap_timer()
...
+	u32 split_time = get_cur_time(); 
+
+	if (curr_index < 20){ // 20 is a magic number that corresponds to the length of the split_times array
+		split_times[curr_index] = split_time; 
+		if (curr_index == 0){
+			segment_times[0] = split_times[0] - main_timer; 
+		}
+		else {
+			segment_times[curr_index] = split_times[curr_index] - split_times_pause_lengths[curr_index] - split_times[curr_index - 1]; 
+		}
```
HeartPiece44: Bug: this doesn't actually account for skipping cutscenes (outside of the first segment) since those increment the main_timer, but not the pause time, and every segment after the first is a relative comparison. Suggestion: make each split `get_cur_time() - main_timer` and then `segment_times[curr_index] = split_times[curr_index] - split_times[curr_index - 1]` rather than having to deal with the pause array at all - feel free to do it differently if you have something in mind for the separate pause lengths array though!

include/p2gz/timer.h
```diff
@@ -93,13 +119,25 @@ struct Timer {
...
+	u32 sub_timer;   // sublevel timer starting point
 	u32 skip_timer;  // for offsets because of P2GZ toggles
 	u32 pause_timer; // for when gz menu is open
 
 	u32 navi_swap_timer; // for measuring captain swap times
 
+	char curr_index; // tracks current active index of split_times
+	u32 split_times[20]; // tracks timestamps of trigger events 
+	u32 segment_times[20]; // tracks the segment times to display on screen 
+	u32 split_times_pause_lengths[20]; // position i tracks the pause length between split time i-1 and split time i
```
HeartPiece44: Bit of a nit: could we make these magic numbers (20) into a `#define MAX_SPLIT_TIMES (20)` or something, so we can change them in the future if we want/makes it clear what they mean? Also involves changing the 20s in timer.cpp as well.

src/p2gz/timer.cpp
```diff
@@ -76,6 +98,74 @@ void Timer::draw()
...
+				}
+
+				if (draw_best_times_enabled && draw_comparisons_enabled && (best_segments[i] > 0)){
+					// fixed offset so it produces aligned columns 
+					f32 sub_offset = 120.0f;
+					bool isGreen = segment_times[i] < best_segments[i]; 
+					if (!isGreen){
+						// alternative: calc_time(best_segments[i], segment_times[i])
+						// I don't like this since these are not the start and endpoint of a period of elapsed time
+						seg_c = calc_time(0, segment_times[i] - best_segments[i]);
```
HeartPiece44, line 142: Maybe outside of scope, so feel free to kick it down the road/ignore it: there's nothing really binding the segments to the settings that you're using to split on. That is, I could run the split with gates and treasures on, then run it with only enemies killed on, and it will still do the comparison as if they're fair. Not the biggest issue at present maybe (don't be an idiot, reset your splits if you change your settings) but maybe more of an issue for comparisons later, especially if we save to mem card.

src/p2gz/timer.cpp
```diff
@@ -76,6 +98,74 @@ void Timer::draw()
...
+					else {
+						seg_c = calc_time(0, best_segments[i] - segment_times[i]);
+						JUtility::TColor green_color(30, 205, 30, 204); 
+						COLOR(green_color);
+					}
+					if (seg_c.minutes > 0){
+						j2d.print(x + sub_offset, starting_seg_offset + (i * 16), isGreen ? "-%ld:%.2ld.%.1ld" : "+%ld:%.2ld.%.1ld", seg_c.minutes, seg_c.seconds, seg_c.tenths);
+					}
+					else{
+						j2d.print(x + sub_offset, starting_seg_offset + (i * 16), isGreen ? "-%.2ld.%.1ld" : "+%.2ld.%.1ld", seg_c.seconds, seg_c.tenths);
```
HeartPiece44: Mega insane nit: being two seconds ahead of the best split time will make this say -02.0, rather than -2.0. Not sure we care enough to adjust that but flagging it anyway in case it will also piss you off to leave it.

src/p2gz/timer.cpp
```diff
@@ -332,6 +431,56 @@ void Timer::cancel_navi_swap_timer()
...
+		}
+		curr_index++;
+	}
+}
+
+
+void Timer::reset_best_segments(){
+	if (!draw_best_times_enabled) {
+		return;
+	}
```
HeartPiece44: I think this (and the guard in the function below) maybe don't make sense? I think you should expect these to wipe your best split times, regardless of if you've toggled on the timer or not tbh, but feel free to disagree and ignore.

src/p2gz/timer.cpp
```diff
@@ -332,6 +431,56 @@ void Timer::cancel_navi_swap_timer()
...
+	for (int i = 0; i < 20; i++){
+		best_segments[i] = 0; 
+	}
+}
+
+// resets all the split/segment time related stuff
+void Timer::reset_split_times(){
+	if (!segment_timer_enabled){
+		return;
+	}
```
HeartPiece44: See comment in `reset_best_segments` - think this guard could be removed for more intuitive functionality but open to being wrong on that.

src/p2gz/timer.cpp
```diff
@@ -23,17 +27,35 @@ Timer::Timer()
...
     , FS_map_flag(false)
     , in_freecam_mode(false)
+	, segment_timer_enabled(true) // controls whether or not the segment timer is turned on 
+	, draw_best_times_enabled(false)
+	, split_on_captain_swap(true)
+	, split_on_gate_seg(false)
+	, split_on_gate_down(true)
+	, split_on_bag_crush(true)
+	, split_on_poison_demo(true)
+	, split_on_carry(false)
```
HeartPiece44: Question: what do we actually want the default out-of-the-box behaviour to be, do we want the segment timer on by default, and it split on mysterious things the user hasn't chosen, or should a user have to toggle it on and be confronted with choosing the settings while doing so? My gut says maybe it should be off, but it might also make people less aware of its existence, so more just checking that you've considered it (you probably have)

HeartPiece44 (changes requested):
This is such an insanely cool feature, and I think people will love it - also thank you for fixing the treasure collection bug while you're here! Noted and appreciated. The backbone of the segment timer seems good, I've just got a few bugs/nits/questions, feel free to shut down most of them, but I think some behaviour (accounting for cutscene skipping, bug around treasure trigger) needs to get updated before we ship it.

src/p2gz/timer.cpp
JpepWasTaken: You're right and you have no idea how annoyed I am to have missed this LOL

jonathonbrady:
Can you run `clang-format` on the files you changed in this PR? There are a lot of minor formatting issues throughout (e.g., missing spaces before opening braces).

src/p2gz/timer.cpp
JpepWasTaken: I have indeed, and yeah I settled on having it on by default for marketing value. I don't think everyone will fully explore submenus or fully read documentation so I am hoping it nudges them towards the feature. I think that works well enough with just splitting on swaps and gate destruction though, I'll just keep those 2.

HeartPiece44 (approved):
Thanks for fixing things up - looks great! We can tweak the fixed-width font stuff later on along with some other prints. 🚢

PR #293, opened 2026-06-23
All presets now have correct treasure counts and pokos

HeartPiece44:
All presets in both PoD and AT have correct treasures flagged as collected, meaning the AT cutscene will trigger when its conditions are met, even when warping to a late-game preset. Treasures are also "kept" correctly on retrying a segment (those collected before where you're resetting to), and others are reset.

All presets in both PoD and AT also have correct pokos recorded, meaning the PoD cutscene will trigger in both categories correctly. 

Presets also now differentiate based on sublevel, for the purposes of pokos and treasure counts, without increasing the number of presets. This makes the internals of presets a bit more complicated, as they require sublevel-specific deltas, but it keeps the interface as simple as possible for players, so probably worth it.

At present, these are US pokos for both, just to get something in - a future pass will need to implement a method to split them for PAL and JP for both categories.


include/p2gz/warp.h
```diff
@@ -89,6 +89,7 @@ struct Warp {
 	bool applying_generators;
 	bool needs_generator_cache_reconstruction; // set when a preset is applied, consumed when cache is rewritten
 	bool do_egate_parse_only_load;             // skip archive/model loading in ItemDengekiGate::Mgr ctor during cache reconstruction
+	bool reset_in_place;                       // segment/sublevel replay; rebuild only the current area
```
jonathonbrady: Nit: it is unclear to me what `reset_in_place` means without consulting this comment. Maybe `only_rebuild_current_area`?

jonathonbrady (approved):

PR #296, opened 2026-06-23
Implement race/speedrun mode

HeartPiece44:
This implements a "race" or speedrun mode, accessible from the P2GZ menu. Players can set their category (PoD or AT) and the ending will be autodetected. Players can start from a fresh file, or from a common later point (SH1/race file for PoD, or WW for AT). Players can either play with vanilla randomised seeds, or with all floors set to the same (enterable) seed, for set seed races. The GZ menu is disabled during race mode.

Two timers are displayed through the run in the top-left: the RTA timer, which runs the whole time, including during pause menus; and the IGT timer, which pauses during loads (as detected by when the game is in LoadState), effectively implementing a load-removed timer.

Race mode overrides the soft reset combo (X-B-Start) to cause it to trigger a retry-segment instead, warping the player back to the last segment that would've been recorded, effectively the last save point. A new combo (hold L+R+Z+Start for ~ 2s) is implemented to exit race mode, since the menu is disabled - this causes a proper soft reset and disables race mode.

image

image

Some fun stats are shown on the black screen after the run (since THPs are removed), including final times, numbers of resets, total pikmin killed, total enemies killed, and number of throws.


src/p2gz/RaceMode.cpp
```diff
@@ -0,0 +1,538 @@
...
+	}
+
+	PresetPreview* preset = p2gz->preset_mgr->suggested_preset(dest, category);
+	if (!preset) {
+		// shouldn't happen (the route always has these presets), but don't warp with no preset
+		OSReport("[P2GZ] race mode: no preset for start point, falling back to fresh file\n");
+		at_fresh_start = true;
+		begin_fresh_file();
+		return;
+	}
```
jonathonbrady: If this shouldn't happen, we should panic instead of silently failing so it's easier to debug.

src/p2gz/RaceMode.cpp
```diff
@@ -0,0 +1,538 @@
...
+{
+	if (!active || finished) {
+		return;
+	}
+
+	// make sure our ending matches the category we set when we started
+	const bool match = (category == AT) ? is_all_treasures : !is_all_treasures;
+	if (!match) {
+		return;
+	}
```
jonathonbrady, line 268: Is this simpler?
```c++
if ((category == AT) != is_all_treasures) {
    return;
}
```

src/p2gz/RaceMode.cpp
```diff
@@ -0,0 +1,538 @@
...
+	} else {
+		j2d.print(decimal_x - j2d.getWidth("%ld:%.2ld", m, s), z, "%ld:%.2ld", m, s);
+	}
+	j2d.print(decimal_x, z, ".%.1ld", t);
+}
+
+void RaceMode::draw_controls_hint()
+{
+	// we lock the menu, so we should let players know how to like. get out of that.
+	// future improvement could be making this toggleable somehow, so experienced players can turn it off
```
jonathonbrady, line 437: Can we address this in v1? If there's no obvious button combo to turn it off, maybe have it fade out after a fixed amount of time, or only draw it during, e.g., sublevel load? Alternatively, we already show these combos in the race mode menu, so maybe we don't need them drawn during gameplay. I just think it might get annoying to have these on screen for the entire duration of a race.

src/p2gz/gzmenu.cpp
```diff
@@ -63,6 +63,13 @@ void GZMenu::init_menu()
 			->push(new PresetMenuOption(new Delegate2<Warp, PresetPreview*, int>(p2gz->warp, &Warp::set_preset)))
 			->push(new PerformActionMenuOption("go", new Delegate<Warp>(p2gz->warp, &Warp::do_warp)))
 		))
+		->push(new OpenSubMenuOption("race", (new ListMenu())
+			->push(new RadioMenuOption("category", new Delegate1<RaceMode, size_t>(p2gz->race_mode, &RaceMode::set_category)))
+			->push(new RadioMenuOption("start point", new Delegate1<RaceMode, size_t>(p2gz->race_mode, &RaceMode::set_start_point)))
+			->push(new HexInputOption("seed", "random", new Delegate1<RaceMode, u32>(p2gz->race_mode, &RaceMode::set_seed), new Delegate<RaceMode>(p2gz->race_mode, &RaceMode::set_random_seed)))
+			->push(new RadioMenuOption("treasure region", new Delegate1<Localization, size_t>(p2gz->localization_op, &Localization::set_treasure_region)))
+			->push(new PerformActionMenuOption("Start run!", new Delegate<RaceMode>(p2gz->race_mode, &RaceMode::start_run)))
```
jonathonbrady: Unfortunately, I hate fun, and this is stylistically inconsistent. Should be `start run`, or just `start`.

jonathonbrady (changes requested):
Very cool overall, just a few minor things.

src/p2gz/RaceMode.cpp
HeartPiece44: Yeah agreed - my only hesitation was trapping people in race mode with no gz menu to cancel things out of, so went a bit heavy on the "make sure the people KNOW about the new combo" - I think we could definitely get away with it just coming up during sublevel loads, and only the abort combo. The softreset override will do what someone running normally would expect but faster, so i think that one's fine to leave to muscle memory? Either that or I can put both just during loads, either or. Will do though.

HeartPiece44:
Updated and also rebased to address some merge conflicts - the button combo hint text now only displays during load transitions, and all other comments are addressed/adjusted, thank you!

jonathonbrady (approved):

PR #297, opened 2026-06-24
Rework warp menu preset selection/suggestion system

HeartPiece44:
Reworks the preset selection system under the warp menu, so players can choose other (actually relevant) presets for an area more easily with dpad-left and dpad-right. Pressing A on the preset selection still brings up the full list to choose from, so no functionality is lost. The internal preset-suggestion system is overhauled to a table structure, with each area/cave/category combo having a list of relevant presets to cycle through.

This now requires players to choose a category on the main warp menu (defaulting to PoD), which feels more intuitive than only choosing through the preset lists. Players can still select category: no preset to warp with current squad.

This also changes the default warp menu option to VoR > Above Ground > PoD > Day 1, rather than the weird EC1 choice it is at the moment.

image


jonathonbrady (approved):
Yeah this is way better. Only thing is that preset icons go off the right side of the screen more often now. Can we put them on their own line (either now if that's easy or in a later release if it takes more than 5 minutes of effort)?

HeartPiece44:
Agreed and implemented :)

PR #303, opened 2026-06-27
Disable skipping breadbug cutscenes until it's been damaged

HeartPiece44:
If "skippable cutscenes" is turned on in settings, players can get in the habit of mashing start to skip the cutscene during gameplay. However, doing so for cutscenes involving a breadbug results in them not being damaged successfully. This makes it so cutscenes involving breadbugs attached the the target treasure are locked from skipping, with the lock being removed once the damage is applied to the breadbug successfully (when it hits the ground).

This does cause some deviation from vanilla behaviour if skipped, since the breadbug would normally be active during the cutscene, and would start seeking out a new target - skipping interrupts this. This only really matters on GK6 during PoD, so the decision is just left up to the player on whether to skip and deviate from the new pathfinding behaviour, or not - it just prevents them from accidentally doing so before they can think it through. I think this is a good balance between preventing mistakes and not taking away player agency.

### DIFF
--- a/include/p2gz/SkippableCS.h
+++ b/include/p2gz/SkippableCS.h
@@ -1,9 +1,12 @@
 #ifndef _GZ_SKIPPABLECS_H
 #define _GZ_SKIPPABLECS_H
 
+#include <types.h>
+
 // predeclarations
 namespace Game {
 struct Creature;
+struct EnemyBase;
 struct MovieConfig;
 } // namespace Game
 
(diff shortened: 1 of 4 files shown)


JpepWasTaken (changes requested):
In-game, you are just totally prevented from skipping any treasure cutscene if a breadbug is attached, it doesn't make it skippable late enough into the cutscene.

It also crashes on GK6 if you try to reset the sublevel (after collecting a treasure I think but needs more thorough testing).

HeartPiece44:
Thanks for the catch - did cleanup on the code after testing and broke it, serves me right. Should work as intended now, sorry!

As for the crash, that's unrelated (and already on main) but the fix for it is in #312.

JpepWasTaken (approved):
Works exactly as expected, good job :)

PR #311, opened 2026-06-28
Make crash log appear much faster

HeartPiece44:
Currently on both console and Dolphin, the crash log takes forever to populate, as the game tries to re-parse the 4.2MB map file to resolve every single address line. This adds a Python script that makes a static lookup table out of the map file, and reroutes the crash handler to do a binary search on it instead when resolving addresses. This reduces the time the crash log takes to come up by a lot, and should make it less insane for people to share crash logs with us.

https://github.com/user-attachments/assets/4574160e-da95-4413-a44c-6680151ba008

### DIFF
--- a/build.py
+++ b/build.py
@@ -168,6 +168,9 @@ if args.map:
     with open("root/files/pikmin2UP.map", "w") as f:
         f.writelines(lines)
 
+    # build the custom compact crash-handler symbol table from the map
+    subprocess.run("python3 tools/gen_crash_symbols.py", shell=True)
+
 
 print(f"Done! Build took {round(time.time() - start_time, 2)}s")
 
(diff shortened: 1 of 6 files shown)


jonathonbrady:
This looks like a Claude one-shot. Can you remove some of the LLMisms? I am also slightly sad that we are shelling out to `python3` subprocesses in our build script, but we already do that for `configure.py`, so it can be cleaned up later.

HeartPiece44:
Definitely had some Claude help - I'd deslopped a fair bit, but I was clearly tired when I did it, so have deslopped it more - lemme know if there's anything more specific to clean up or improve though and I'll tackle it again. To test out, go to an AG and toggle a treasure on and off for a bit.

Noted re: the python3 subprocesses, but unfortunately that was me just fitting in with the style we already had. Agree we should clean it up later as a whole though.

jonathonbrady (approved):

PR #314, opened 2026-06-29
More race mode tweaks

### LINKED ISSUE #310: Vanilla race feature request
As it is right now, to select all the settings that correspond to a vanilla race (unskippable cutscenes...) one needs to go through all the other menus to set these options. It would be nice if they were all accessible from the race mode screen.

HeartPiece44:
This fixes #310 by adding in the skippable cutscenes and save prompt skipping settings to the race mode menu, to make it more clear what settings are turned on when starting. It also resets all other settings to vanilla when starting race mode, so nothing custom is on.

This also includes some other QoL fixes for race mode, including capturing a few more loads that were missed (primarily the end-of-day results into world map load) since they use custom substates, rather than the global LoadState. It also makes sure the end of race results/stats actually show legibly over the top of the ending results.


JpepWasTaken:
lgtm!

JpepWasTaken (approved):

PR #315, opened 2026-06-30
fix camera position in find blue onion cutscene

jonathonbrady:
Our `moviePlayer.cpp` wasn't equivalent for some reason: https://github.com/projectPiki/pikmin2/blob/main/src/sysGCU/moviePlayer.cpp#L636


JpepWasTaken (approved):
Tested it in game and it works, LGTM!

PR #316, opened 2026-06-30
Segment timer fixes 

JpepWasTaken:
Quick segment timer fixes: uses HP's race mode timer display function, added guardrail checks to make intended behaviour clearer, turned off by default, splits on cave entry textbox.


src/plugProjectKandoU/navi.cpp
```diff
@@ -1926,11 +1926,14 @@ void Navi::update()
...
+	// @P2GZ: freecam and segment timer
+	// don't open hole menus when enabling freecam + split when pressing A near hole 
 	// if (hole && mController1 && mController1->getButtonDown() & Controller::PRESS_A) {
 	if (hole && mController1 && mController1->getButtonDown() & Controller::PRESS_A && !p2gz->freecam->is_enabled()) {
 		gameSystem->mSection->openCaveMoreMenu(hole, mController1);
+		if (p2gz->timer->get_segment_timer_enabled() && p2gz->timer->get_segment_timer_enabled() && p2gz->timer->split_on_cave_enter) {
+			p2gz->timer->add_split_times(); 
+		}
 		return;
 	}
```
HeartPiece44, line 1938: This is good, but it only works for the "next floor" hole in caves - if you want it to work for above grounds as well, you'll need to do the same fix for the next block down in this file (the one for `ItemCave::Item* cave`). If you want it to also work for geysers (which I think would be sensible) that would be the block after that as well.

src/plugProjectKandoU/navi.cpp
```diff
@@ -1926,11 +1926,14 @@ void Navi::update()
...
 	ItemHole::Item* hole = checkHole();
 
-	// @P2GZ: freecam
-	// don't open hole menus when enabling freecam
+	// @P2GZ: freecam and segment timer
+	// don't open hole menus when enabling freecam + split when pressing A near hole 
 	// if (hole && mController1 && mController1->getButtonDown() & Controller::PRESS_A) {
 	if (hole && mController1 && mController1->getButtonDown() & Controller::PRESS_A && !p2gz->freecam->is_enabled()) {
 		gameSystem->mSection->openCaveMoreMenu(hole, mController1);
+		if (p2gz->timer->get_segment_timer_enabled() && p2gz->timer->get_segment_timer_enabled() && p2gz->timer->split_on_cave_enter) {
```
HeartPiece44: I think you've got an extra `get_segment_timer_enabled()` that could get removed?

src/p2gz/timer.cpp
```diff
@@ -120,20 +123,23 @@ void Timer::draw()
 				}
 
 				if (seg_c.minutes > 0) {
-					j2d.print(x, starting_seg_offset + (i * 16), "%ld:%ld.%ld", seg_c.minutes, seg_c.seconds, seg_c.tenths);
+					j2d.print(decimal_x - j2d.getWidth("%ld:%2ld.%ld", seg_c.minutes, seg_c.seconds, seg_c.tenths), starting_seg_offset + (i * 16),
```
HeartPiece44: Bug: when a segment time is say, 1 minute 8 seconds, 500 ms, the display will show 1: 8.5 rather than 1:08.5 - I think these `%ld:%2ld.%1d` should be `%ld:%.2ld.%.1ld` like on line 103. This also goes for the other spots that use this below.

src/p2gz/timer.cpp
```diff
@@ -120,20 +123,23 @@ void Timer::draw()
...
 
 				if (draw_best_times_enabled && (best_segments[i] > 0)) {
 					// fixed offset so it produces aligned columns
 					f32 sub_offset = 60.0f;
 					seg_c          = calc_time(0, best_segments[i]);
 					if (seg_c.minutes > 0) {
-						j2d.print(x + sub_offset, starting_seg_offset + (i * 16), "%ld:%ld.%ld", seg_c.minutes, seg_c.seconds,
-						          seg_c.tenths);
+						j2d.print(decimal_x + sub_offset - j2d.getWidth("%ld:%2ld.%ld", seg_c.minutes, seg_c.seconds, seg_c.tenths),
+						          starting_seg_offset + (i * 16), "%ld:%2ld.%ld", seg_c.minutes, seg_c.seconds, seg_c.tenths);
```
HeartPiece44: See above re: missing 0 in seconds

src/p2gz/timer.cpp
```diff
@@ -153,11 +159,14 @@ void Timer::draw()
 						COLOR(green_color);
 					}
 					if (seg_c.minutes > 0) {
-						j2d.print(x + sub_offset, starting_seg_offset + (i * 16), isGreen ? "-%ld:%ld.%ld" : "+%ld:%ld.%ld", seg_c.minutes,
-						          seg_c.seconds, seg_c.tenths);
+						j2d.print(
+						    decimal_x + sub_offset
+						        - j2d.getWidth(isGreen ? "-%ld:%2ld.%ld" : "+%ld:%ld.%ld", seg_c.minutes, seg_c.seconds, seg_c.tenths),
+						    starting_seg_offset + (i * 16), isGreen ? "-%ld:%2ld.%ld" : "+%ld:%ld.%ld", seg_c.minutes, seg_c.seconds,
+						    seg_c.tenths);
```
HeartPiece44: See above re: missing 0 in seconds.

HeartPiece44 (changes requested):
Nice, thanks for fixing this up! Much appreciated. Noticed a couple bugs still, should all be quick fixes I think, then should be good to go.

HeartPiece44 (approved):
Thanks for the fixes - looks good now to me!

PR #318, opened 2026-06-30
Fix some treasure editor crashes

HeartPiece44:
This fixes a few treasure editor-related crashes and bugs:
- Fixes a crash when trying to respawn a previously collected treasure
- Fixes a bug where respawning a treasure over and over would cause the pellet count to fall out of sync with the actual spawned count
- Patches a heap crash when toggling a treasure to collected/uncollected too many times by preventing the player from spawning any more. This resets on area reload. This one is caused by how pellets are destructed in the game, and would require a lot more meddling than I'm comfortable doing right now. If we want to add some warning text or something we can, but I think it's unlikely people will hit the silent limit without them actively fucking around.

### DIFF
--- a/include/p2gz/TreasureEditor.h
+++ b/include/p2gz/TreasureEditor.h
@@ -5,6 +5,10 @@
 #include <p2gz/gzCollections.h>
 #include <Game/Entities/PelletItem.h>
 
+namespace Game {
+struct PelletConfig;
+}
+
 namespace gz {
 
 struct TreasureEditor {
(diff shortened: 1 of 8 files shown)


HeartPiece44:
Have added a couple extra things to this, based on some other reported bugs, should be much more stable hopefully:
- Fixes treasures avoiding the heap guard in caves and continuing to spawn anyway
- Fixes how buried treasures are dealt with, so mysterious extra dig spots don't appear every time you toggle the treasure
- Fixes how the treasure editor operates in caves - now, treasures will properly respawn in the correct location (except for those spawning in enemies, future todo)
- Fixes a crash with menus that can resize on entering/exiting a submenu (due to out-of-bounds indexing)
- Treasure menu now will permanently show "collected" when hitting the heap guard, to properly reflect the game state

JpepWasTaken (dismissed):
Good work, LGTM! Let's get this release cooking.

JpepWasTaken:
Just a couple tiny tiny notes: 
-when the camera is focused on a treasure, the treasure gauge reacts as if you were very close to it. I don't think this is particularly an issue at all but I'm making a note of it 
-the "no memory to spawn treasure" message should maybe get cleared when the player returns to the previous menu (selecting which treasure to edit)

HeartPiece44:
Agree on noting these and fixing them down the track in a later cleanup pass - I think looking into them now is probably more effort than they're worth, and I reckon our time is better spent testing things out imo, unless you feel more strongly.

HeartPiece44:
After some more discussion and your suggestion on using mDoSkipCreateModel, the treasure editor now doesn't crash the game when toggling the treasure, so the callout text has been removed. The gauge fixes are left to a future PR though.

JpepWasTaken (approved):
The beast is slain, good job team!

PR #319, opened 2026-06-30
EB Trainer Fixes

JpepWasTaken:
Add camera inversion toggles for EB trainer, fix bug where the trainer incorrectly labels successes as wrong waypoint failures, adds functionality so pressing L centers the inset cam.


HeartPiece44 (commented):
Nice work! Looks slick, and it's great having the extra functionality, I hadn't realised how much it irked me until I had the option of changing it lol. One suggestion for changes: having the EB camera options just In The Trainer Menu feels a bit weird - maybe worth either having them hidden unless the EB trainer is active, or having a submenu for EB trainer with them in it + start? Choice is up to you, but probably good to sweep them somewhere if we wanna expand the trainer menu with more cool trainers down the line.

HeartPiece44 (approved):

PR #322, opened 2026-07-03
Fix gz menu in % cutscenes

JpepWasTaken:
Fixes a bug where GZ menu could be opened during or before percent cutscenes.

### DIFF
--- a/include/p2gz/Utility.h
+++ b/include/p2gz/Utility.h
@@ -344,6 +344,21 @@ inline bool in_cave_results()
 	return true;
 }
 
+inline bool in_repay_demo()
+{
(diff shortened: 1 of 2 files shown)


HeartPiece44 (approved):
LGTM! 🚢 

Noting for future: I wonder if there's any way for us to allow opening the gz menu during the percent cutscene, but just avoid any nonsense that comes with it - like delaying the percent cutscene until the menu is closed. Waiting the full lockout if you mis-warp somewhere is probably a bit annoying. Works well for now though, and better than people breaking stuff by opening it before the cutscene starts.

PR #328, opened 2026-07-06
Fix wrong copy in Linux patcher

Julian-Wollersberger:
It did create a subdirectory root/sys/sys instead of overwriting the files.

Tested on Ubuntu 26.04


jonathonbrady (approved):
Thanks for fixing this!

Julian-Wollersberger:
Can you merge this? I don't have permissions.

